Issue #79, opened 2011-11-27
Opening non-pyramidal tiled TIFF is excruciatingly slow


bgilbert:
quickhash-1 over TIFF files is defined to include all tiles in the lowest-resolution layer, so if a generic tiled TIFF is not pyramidal, we are computing the hash over all of the image data in the file.  This can cause `openslide_open()` to take 60 seconds or more for large TIFFs.

Arguably, OpenSlide isn't meant for non-pyramidal images.  However, a program doing generic format detection on a file (e.g. VIPS) doesn't know any better.

(In the VIPS case, it's actually even worse.  The VIPS OpenSlide driver does not want to support `generic-tiff` images, since the VIPS TIFF driver can do it better.  But it still has to call `openslide_open()` to check `openslide.vendor`, incurring the quickhash overhead even though we're not going to use OpenSlide for this file.)

Options:
1. Permit deferring quickhash calculation until the attribute is read.  Probably not useful for the other vendors; on the CMU dataset quickhash calculation is always less than half the total `openslide_open()` time.
2. Refuse to open a generic-tiff that is not pyramidal.  This presumes that pyramidal TIFFs will not have perversely large top layers.
3. Others?

PR #157, opened 2015-03-23
Add support for DICOM (aka Supp 145)

malaterre:
This commit adds support for VL Whole Slide Microscopy Image IOD

---

This change is [Reviewable](https://reviewable.io/reviews/openslide/openslide/157)


bgilbert:
Hi Mathieu,

Thanks for the submission, and I apologize for the protracted delay in getting this reviewed.  The submitted code is pretty rough, and would need a lot of cleanup to get it into mergeable shape.  There are two issues in particular that must be fixed before the next iteration:

- There are coding style violations throughout the PR.
- Assertions (using `g_assert`, not `assert`) should only be used to detect programming errors.  Data errors must be reported with GError.  This means that every function that can fail should have a GError out-argument, unless there's a very good reason to do otherwise.

Because this version will likely require significant restructuring, I've reviewed it as an RFC: I pointed out a selection of issues I noticed, but haven't carefully reviewed the code in detail.

---

Reviewed 6 of 6 files at r1.
Review status: all files reviewed at latest revision, 37 unresolved discussions.

---

*[a discussion](https://reviewable.io:443/reviews/openslide/openslide/157#-KQJRDxZIqlUwwwmpxY8:-KQJRDxZIqlUwwwmpxY9:b3m6gjn) (no related file):*
I think the major question is whether we should really be implementing our own DICOM parser, or whether we should use an existing DICOM library instead.  Could you comment on the tradeoffs from your perspective?  The decoder implemented here is already a lot of code, and apparently doesn't handle cases (missing DICOMDIR) that we may need to support.  I'd rather avoid maintaining this complexity in OpenSlide if possible, especially if it requires duplicating code already available elsewhere.  (The only reason we implement our own TIFF parser is that libtiff can't handle NDPI.)

As I understand it, we could use a C++ DICOM library by implementing openslide-decode-dicom in C++, exposing only C interfaces.  We'd need to link OpenSlide using the C++ linker, but as far as I can tell, users of the library wouldn't see any difference.

---

*[a discussion](https://reviewable.io:443/reviews/openslide/openslide/157#-KQw6YrLYiXEUrjET0Ym:-KQJREsHnSMx0pOLNuzA:b-3q3sta) (no related file):*
The API between `decode` and `vendor` is probably at the wrong level.  `openslide-decode-tiff.h`, and in particular `struct _openslide_tiff_level`, is not a good model for DICOM: the DICOM decoder shouldn't need to know what specific pieces of information the vendor driver cares about.  (The TIFF API does things that way because of the large number of vendor drivers which need to store identical information in their level structs.)  The style of API in `openslide-decode-tifflike.h` would be cleaner and more idiomatic: accessor methods on opaque structs that allow the vendor driver to query specific tags.

---

*[a discussion](https://reviewable.io:443/reviews/openslide/openslide/157#-KQJRg2nXIzeHG8GXx99:-KQJRg2o-DyNSHlHd7sL:b-yaketc) (no related file):*
The test suite will need test cases for this format, which means we'll need some redistributable sample data.  I'm not aware of any public sample files which are explicitly redistributable; can you provide some?

---

*[a discussion](https://reviewable.io:443/reviews/openslide/openslide/157#-KQw6mmlrt2mdZAISxL_:-KQJWDx4eci8yj7OEeYz:ba75v5w) (no related file):*
What are the known limitations of the current implementation?

---

*[a discussion](https://reviewable.io:443/reviews/openslide/openslide/157#-KQJlIphM_ULmD62ZYEA:-KQJlIphM_ULmD62ZYEB:bm46h6v) (no related file):*
There is no support for properties and associated images.  This should be added, though perhaps not immediately.

---

*[Makefile.am, line 79 \[r1\]](https://reviewable.io:443/reviews/openslide/openslide/157#-KQJWW9_egdD6EgNmvOF:-KQJWW9_egdD6EgNmvOG:b-wfkh5g) ([raw file](https://github.com/openslide/openslide/blob/ac517ffa5030e851d4373e082e4a7f1bc3d37a7f/Makefile.am#L79)):*
> ```Makefile
> 
> noinst_HEADERS = \
> 	src/openslide-cairo.h \
> ```

Need to add `openslide-decode-dicom.h`.

---

*[src/openslide-decode-dicom.c, line 56 \[r1\]](https://reviewable.io:443/reviews/openslide/openslide/157#-KQJe84uyuc1NYkrWToO:-KQJe84uyuc1NYkrWToP:bv9o0hx) ([raw file](https://github.com/openslide/openslide/blob/ac517ffa5030e851d4373e082e4a7f1bc3d37a7f/src/openslide-decode-dicom.c#L56)):*
> ```C
>  * Because OpenSlide (as of ABI 1.0) only deal with single slide, using the
>  * Study Instance UID is enough to generate a unique quickhash. In the long
>  * term one may want to pad the quickhash with the Series Instance UID.
> ```

This comment should go in the vendor driver, not here.

See quickhash discussion [here](https://lists.andrew.cmu.edu/pipermail/openslide-users/2015-March/001030.html).  Also, we'll want to hash the lowest-resolution pyramid level in case of UID collisions.

---

*[src/openslide-decode-dicom.c, line 74 \[r1\]](https://reviewable.io:443/reviews/openslide/openslide/157#-KQJVHommHOnxkjNVQUd:-KQJVHonKE-Bs3tVpUWS:bjbk0wd) ([raw file](https://github.com/openslide/openslide/blob/ac517ffa5030e851d4373e082e4a7f1bc3d37a7f/src/openslide-decode-dicom.c#L74)):*
> ```C
>  * $ mv Dog_15x15_40x.dcm DOG15_40
>  * $ dcmmkdir --output-file DOGDIR --general-purpose-dvd DOG15_20 DOG15_40
>  * -> Points openslide to the newly created index file: `DOGDIR`
> ```

Is this a scenario likely to be encountered by users?  If so, we should be able to read the files as-is, without requiring an external indexing step.

---

*[src/openslide-decode-dicom.c, line 89 \[r1\]](https://reviewable.io:443/reviews/openslide/openslide/157#-KQJVixS9PIJA7bfkP9T:-KQJVixS9PIJA7bfkP9U:bswm2yk) ([raw file](https://github.com/openslide/openslide/blob/ac517ffa5030e851d4373e082e4a7f1bc3d37a7f/src/openslide-decode-dicom.c#L89)):*
> ```C
> typedef uint32_t tag_t;
> typedef uint16_t vr_t;
> typedef uint32_t vl_t;
> ```

typedefs of primitive types are not useful in C; please remove.

---

*[src/openslide-decode-dicom.c, line 107 \[r1\]](https://reviewable.io:443/reviews/openslide/openslide/157#-KQJX6dtQR5qHZIF0T9X:-KQJX6dtQR5qHZIF0T9Y:bmtmznn) ([raw file](https://github.com/openslide/openslide/blob/ac517ffa5030e851d4373e082e4a7f1bc3d37a7f/src/openslide-decode-dicom.c#L107)):*
> ```C
> #define MAKE_TAG(group,element) (group << 16 | element)
> 
> #if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
> ```

These conditionals should be eliminated and replaced with inline calls to the [glib byte-order macros](https://developer.gnome.org/glib/stable/glib-Byte-Order-Macros.html).

---

*[src/openslide-decode-dicom.c, line 210 \[r1\]](https://reviewable.io:443/reviews/openslide/openslide/157#-KQJXjw1qSy_kUp2ETv1:-KQJXjw1qSy_kUp2ETv2:b-wjvcpe) ([raw file](https://github.com/openslide/openslide/blob/ac517ffa5030e851d4373e082e4a7f1bc3d37a7f/src/openslide-decode-dicom.c#L210)):*
> ```C
> }
> 
> typedef struct
> ```

We don't typedef structs in OpenSlide.

---

*[src/openslide-decode-dicom.c, line 217 \[r1\]](https://reviewable.io:443/reviews/openslide/openslide/157#-KQJfDkF0DTOC1UoOG1s:-KQJfDkF0DTOC1UoOG1t:b-60fwqx) ([raw file](https://github.com/openslide/openslide/blob/ac517ffa5030e851d4373e082e4a7f1bc3d37a7f/src/openslide-decode-dicom.c#L217)):*
> ```C
> } dataelement;
> 
> static inline bool tag_equal_to( const dataelement * de, tag_t tag )
> ```

Let the compiler decide when to inline things, unless profiling indicates otherwise.

---

*[src/openslide-decode-dicom.c, line 283 \[r1\]](https://reviewable.io:443/reviews/openslide/openslide/157#-KQJ_xHX0SUeXD9mj7fm:-KQJ_xHX0SUeXD9mj7fn:b-524kdp) ([raw file](https://github.com/openslide/openslide/blob/ac517ffa5030e851d4373e082e4a7f1bc3d37a7f/src/openslide-decode-dicom.c#L283)):*
> ```C
> static tag_path * create_tag_path(void)
> {
>   tag_path *ptr = (tag_path*)malloc( sizeof(tag_path) );
> ```

We always use `g_slice_new0`/`g_slice_free` when allocating structs, and `g_malloc`/`g_free` when allocating dynamically-sized objects.

---

*[src/openslide-decode-dicom.c, line 801 \[r1\]](https://reviewable.io:443/reviews/openslide/openslide/157#-KQJZ3mjKacS15OSZD2H:-KQJZ3mjKacS15OSZD2I:b-lgcy8t) ([raw file](https://github.com/openslide/openslide/blob/ac517ffa5030e851d4373e082e4a7f1bc3d37a7f/src/openslide-decode-dicom.c#L801)):*
> ```C
> // END DICOM Implementation
> 
> // OpenSlide specific hooks (may use glib)
> ```

The entire file may use glib, and should.

---

*[src/openslide-decode-dicom.c, line 803 \[r1\]](https://reviewable.io:443/reviews/openslide/openslide/157#-KQJkAuZa6QdKnVOXENW:-KQJkAu_q7cqefK97CUI:b-ccuxsg) ([raw file](https://github.com/openslide/openslide/blob/ac517ffa5030e851d4373e082e4a7f1bc3d37a7f/src/openslide-decode-dicom.c#L803)):*
> ```C
> // OpenSlide specific hooks (may use glib)
> 
> struct dicom_info {
> ```

struct definitions should go at the top of the file.

---

*[src/openslide-decode-dicom.c, line 804 \[r1\]](https://reviewable.io:443/reviews/openslide/openslide/157#-KQJg4tq5OVSpKx_JRHV:-KQJg4tq5OVSpKx_JRHW:b-kjv2s8) ([raw file](https://github.com/openslide/openslide/blob/ac517ffa5030e851d4373e082e4a7f1bc3d37a7f/src/openslide-decode-dicom.c#L804)):*
> ```C
> 
> struct dicom_info {
>   int number_of_frames;
> ```

Explicit sizes (`int32_t`, etc.) are preferred.

---

*[src/openslide-decode-dicom.c, line 809 \[r1\]](https://reviewable.io:443/reviews/openslide/openslide/157#-KQJiggg0GLtfPYGKbnL:-KQJigghi_JguDWn6Lwj:bhzdwxu) ([raw file](https://github.com/openslide/openslide/blob/ac517ffa5030e851d4373e082e4a7f1bc3d37a7f/src/openslide-decode-dicom.c#L809)):*
> ```C
>   int total_pixel_mat_cols;
>   int total_pixel_mat_rows;
>   char code_value[16];
> ```

We always allocate strings (because that's the approach preferred by glib functions) rather than using char array members.

---

*[src/openslide-decode-dicom.c, line 816 \[r1\]](https://reviewable.io:443/reviews/openslide/openslide/157#-KQJZIBovaSFnZsJpgow:-KQJZIBovaSFnZsJpgox:b-76lvjh) ([raw file](https://github.com/openslide/openslide/blob/ac517ffa5030e851d4373e082e4a7f1bc3d37a7f/src/openslide-decode-dicom.c#L816)):*
> ```C
> };
> 
> static const char *trimwhitespace(char *str)
> ```

glib has a function for this.

---

*[src/openslide-decode-dicom.c, line 833 \[r1\]](https://reviewable.io:443/reviews/openslide/openslide/157#-KQw-o8KBjo947Q4XSOh:-KQw-o8KBjo947Q4XSOi:bj7yc6q) ([raw file](https://github.com/openslide/openslide/blob/ac517ffa5030e851d4373e082e4a7f1bc3d37a7f/src/openslide-decode-dicom.c#L833)):*
> ```C
>     GSList * list = (GSList*)ds->data;
>     char buf[512];
>     fread( buf, 1, len, stream );
> ```

Possible overrun?

---

*[src/openslide-decode-dicom.c, line 835 \[r1\]](https://reviewable.io:443/reviews/openslide/openslide/157#-KQw-Mlhl26zvbgAK19u:-KQw-Mlhl26zvbgAK19v:bmjcqn0) ([raw file](https://github.com/openslide/openslide/blob/ac517ffa5030e851d4373e082e4a7f1bc3d37a7f/src/openslide-decode-dicom.c#L835)):*
> ```C
>     fread( buf, 1, len, stream );
>     buf[len] = 0;
>     // change windows style backslash into UNIX style forward slash
> ```

Won't this break on Windows?

---

*[src/openslide-decode-dicom.c, line 864 \[r1\]](https://reviewable.io:443/reviews/openslide/openslide/157#-KQJjMy55x_B8Wn5PXl5:-KQJjMy6SycCouaIieaB:b-xmlhd5) ([raw file](https://github.com/openslide/openslide/blob/ac517ffa5030e851d4373e082e4a7f1bc3d37a7f/src/openslide-decode-dicom.c#L864)):*
> ```C
>     uint32_t ul;
>     memcpy( &us, buf, sizeof us );
>     memcpy( &ul, buf, sizeof ul );
> ```

See `read_uint()` in openslide-decode-tifflike.c.  We have several such static functions throughout the library; probably they should be consolidated into an openslide-decode-data.c.  (But that's not necessarily something you need to do.)

---

*[src/openslide-decode-dicom.c, line 877 \[r1\]](https://reviewable.io:443/reviews/openslide/openslide/157#-KQw0Q9HAuOsveD935x3:-KQw0Q9HAuOsveD935x4:bhrgols) ([raw file](https://github.com/openslide/openslide/blob/ac517ffa5030e851d4373e082e4a7f1bc3d37a7f/src/openslide-decode-dicom.c#L877)):*
> ```C
>         break;
>       case MAKE_TAG(0x0028,0x0008):
>         di->number_of_frames = atoi( buf );
> ```

Use `g_ascii_strtoll`.  `atoi` doesn't report errors.

---

*[src/openslide-decode-dicom.c, line 893 \[r1\]](https://reviewable.io:443/reviews/openslide/openslide/157#-KQJgLJlxtfMZOC-A_zn:-KQJgLJlxtfMZOC-A_zo:b-vy0jas) ([raw file](https://github.com/openslide/openslide/blob/ac517ffa5030e851d4373e082e4a7f1bc3d37a7f/src/openslide-decode-dicom.c#L893)):*
> ```C
>       default:
>         // programmer error, need to handle attribute
>         assert(0);
> ```

Use `g_assert_not_reached()` instead of `assert(0)`.

---

*[src/openslide-decode-dicom.c, line 949 \[r1\]](https://reviewable.io:443/reviews/openslide/openslide/157#-KQORW9gAncAoL34Ggxh:-KQORW9gAncAoL34Ggxi:by2en9k) ([raw file](https://github.com/openslide/openslide/blob/ac517ffa5030e851d4373e082e4a7f1bc3d37a7f/src/openslide-decode-dicom.c#L949)):*
> ```C
>   // Setup reader to parse DICOMDIR instances
>   assert( instance->ds.tps->nsets == 0 );
>     {
> ```

Why is there a block here?

---

*[src/openslide-decode-dicom.c, line 1063 \[r1\]](https://reviewable.io:443/reviews/openslide/openslide/157#-KQw1eYnwTMAJZ1rdhJE:-KQw1eYnwTMAJZ1rdhJF:b-pvfgh0) ([raw file](https://github.com/openslide/openslide/blob/ac517ffa5030e851d4373e082e4a7f1bc3d37a7f/src/openslide-decode-dicom.c#L1063)):*
> ```C
>     || !read_dataset( &instance->ds, instance->stream ) )
>     {
>     return false;
> ```

Does not set `*err`.

---

*[src/openslide-decode-dicom.c, line 1103 \[r1\]](https://reviewable.io:443/reviews/openslide/openslide/157#-KQJUVyCqp65-Ac4tPjQ:-KQJUVyCqp65-Ac4tPjR:bj3ih3w) ([raw file](https://github.com/openslide/openslide/blob/ac517ffa5030e851d4373e082e4a7f1bc3d37a7f/src/openslide-decode-dicom.c#L1103)):*
> ```C
>     strncpy(dicoml->hash, di.study_instance_uid, di.study_instance_uid_len );
>     dicoml->hash[di.study_instance_uid_len] = 0;
>     dicoml->image_format = FORMAT_JPEG;
> ```

Unused; should be removed (along with the struct field and the enum definition).

---

*[src/openslide-decode-dicom.c, line 1112 \[r1\]](https://reviewable.io:443/reviews/openslide/openslide/157#-KQJkTQYRuBILTq8bhso:-KQJkTQZBf7COz5-yhEo:bklbe3m) ([raw file](https://github.com/openslide/openslide/blob/ac517ffa5030e851d4373e082e4a7f1bc3d37a7f/src/openslide-decode-dicom.c#L1112)):*
> ```C
> bool _openslide_dicom_is_dicomdir(const char *filename, GError **err)
> {
>   FILE *stream = _openslide_fopen(filename, "rb", err);
> ```

Does not check for failure.

---

*[src/openslide-decode-dicom.h, line 33 \[r1\]](https://reviewable.io:443/reviews/openslide/openslide/157#-KQJmO2YeXUXtg2owRgU:-KQJmO2YeXUXtg2owRgV:b-knzj8d) ([raw file](https://github.com/openslide/openslide/blob/ac517ffa5030e851d4373e082e4a7f1bc3d37a7f/src/openslide-decode-dicom.h#L33)):*
> ```objc
> 
> /* DICOM container support for VL Whole Slide Microscopy Image Storage (1.2.840.10008.5.1.4.1.1.77.1.6) */
> /* Thread-safe. */
> ```

`_openslide_dicom_readindex()` is not thread-safe.

---

*[src/openslide-private.h, line 136 \[r1\]](https://reviewable.io:443/reviews/openslide/openslide/157#-KQOOBMsG6EOkJR7sQyu:-KQOOBMsG6EOkJR7sQyv:bkycap1) ([raw file](https://github.com/openslide/openslide/blob/ac517ffa5030e851d4373e082e4a7f1bc3d37a7f/src/openslide-private.h#L136)):*
> ```objc
> extern const struct _openslide_format _openslide_format_trestle;
> extern const struct _openslide_format _openslide_format_ventana;
> extern const struct _openslide_format _openslide_format_dicom_wsmis;
> ```

Alphabetical order, please.

---

*[src/openslide-vendor-dicom.c, line 52 \[r1\]](https://reviewable.io:443/reviews/openslide/openslide/157#-KQOS4y2Z8Oz3219iZiw:-KQOS4y2Z8Oz3219iZix:b-8gf46q) ([raw file](https://github.com/openslide/openslide/blob/ac517ffa5030e851d4373e082e4a7f1bc3d37a7f/src/openslide-vendor-dicom.c#L52)):*
> ```C
> static void destroy(openslide_t *osr) {
>   struct dicom_wsmis_ops_data *data = osr->data;
>   g_slice_free(struct dicom_wsmis_ops_data, data);
> ```

Leaks `datafile_paths`.

---

*[src/openslide-vendor-dicom.c, line 57 \[r1\]](https://reviewable.io:443/reviews/openslide/openslide/157#-KQOSI2MkMvZmXnzoJo2:-KQOSI2MkMvZmXnzoJo3:b-f44g3y) ([raw file](https://github.com/openslide/openslide/blob/ac517ffa5030e851d4373e082e4a7f1bc3d37a7f/src/openslide-vendor-dicom.c#L57)):*
> ```C
>     struct level *l = (struct level *) osr->levels[i];
>     _openslide_grid_destroy(l->grid);
>     g_slice_free(struct level, l);
> ```

Leaks tile array.

---

*[src/openslide-vendor-dicom.c, line 155 \[r1\]](https://reviewable.io:443/reviews/openslide/openslide/157#-KQJkh3jBcJA6W-6HlqX:-KQJkh3kEE7inlf9RpkB:bphbbpa) ([raw file](https://github.com/openslide/openslide/blob/ac517ffa5030e851d4373e082e4a7f1bc3d37a7f/src/openslide-vendor-dicom.c#L155)):*
> ```C
> 
>   // verify existence
>   if (!g_file_test(filename, G_FILE_TEST_EXISTS)) {
> ```

Redundant; `_openslide_dicom_is_dicomdir()` should catch this.

---

*[src/openslide-vendor-dicom.c, line 163 \[r1\]](https://reviewable.io:443/reviews/openslide/openslide/157#-KQJkmEy2OL2pPNWDmwp:-KQJkmEy2OL2pPNWDmwq:b-qtvw8e) ([raw file](https://github.com/openslide/openslide/blob/ac517ffa5030e851d4373e082e4a7f1bc3d37a7f/src/openslide-vendor-dicom.c#L163)):*
> ```C
>   // ensure DICOM is DICOMDIR instance
>   if (!_openslide_dicom_is_dicomdir(filename, err)) {
>     g_set_error(err, OPENSLIDE_ERROR, OPENSLIDE_ERROR_FAILED,
> ```

Sets error on top of existing error, which is forbidden by the GError rules.

---

*[src/openslide-vendor-dicom.c, line 189 \[r1\]](https://reviewable.io:443/reviews/openslide/openslide/157#-KQJhnjMYnKbnKeKDelr:-KQJhnjQ3Gf0cM2J0_m7:b-dmkzya) ([raw file](https://github.com/openslide/openslide/blob/ac517ffa5030e851d4373e082e4a7f1bc3d37a7f/src/openslide-vendor-dicom.c#L189)):*
> ```C
>   char *dirname = NULL;
>   // get directory from filename
>   char *end = strrchr(filename, '/');
> ```

Use `g_path_get_dirname()`.

---

*[src/openslide-vendor-dicom.c, line 193 \[r1\]](https://reviewable.io:443/reviews/openslide/openslide/157#-KQOSv1YbtAFj9HbsD8s:-KQOSv1YbtAFj9HbsD8t:b-74knfd) ([raw file](https://github.com/openslide/openslide/blob/ac517ffa5030e851d4373e082e4a7f1bc3d37a7f/src/openslide-vendor-dicom.c#L193)):*
> ```C
>     dirname = g_strndup(filename, end - filename );
>   else
>     dirname = g_strndup(".", 1);
> ```

Leaked.

---

*[src/openslide-vendor-dicom.c, line 235 \[r1\]](https://reviewable.io:443/reviews/openslide/openslide/157#-KQJnnvq8xFGiL9n8A9i:-KQJnnvq8xFGiL9n8A9j:b4igyyw) ([raw file](https://github.com/openslide/openslide/blob/ac517ffa5030e851d4373e082e4a7f1bc3d37a7f/src/openslide-vendor-dicom.c#L235)):*
> ```C
>     // add to array
>     if( !dicoml->is_icon )
>       g_ptr_array_add(level_array, l);
> ```

Otherwise leak it?

---

*[src/openslide-vendor-dicom.c, line 269 \[r1\]](https://reviewable.io:443/reviews/openslide/openslide/157#-KQJS7ghH-FudKCxYRPu:-KQJS7giIMBoX80DRfDC:b-ce4eaw) ([raw file](https://github.com/openslide/openslide/blob/ac517ffa5030e851d4373e082e4a7f1bc3d37a7f/src/openslide-vendor-dicom.c#L269)):*
> ```C
> }
> 
> const struct _openslide_format _openslide_format_dicom_wsmis = {
> ```

Is there any reason not to drop `wsmis` and just refer to the format as `dicom`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openslide/openslide/157#-:-KQJruPTh6ydNXvf6tuZ:b-r2liet)*

bgilbert:
Review status: all files reviewed at latest revision, 41 unresolved discussions.

---

*[src/openslide-decode-dicom.c, line 917 at r1](https://reviewable.io:443/reviews/openslide/openslide/157#-KT7mFyWO-hxEwb2qtqc:-KT7mFyWO-hxEwb2qtqd:b-5yl7qo) ([raw file](https://github.com/openslide/openslide/blob/ac517ffa5030e851d4373e082e4a7f1bc3d37a7f/src/openslide-decode-dicom.c#L917)):*
> ```C
>   tiles = di->tiles;
>   assert( tiles );
>   assert( di->current_tile_num < di->number_of_frames );
> ```

Are we likely to encounter files that use multiple fragments per frame?

---

*[src/openslide-decode-dicom.c, line 956 at r1](https://reviewable.io:443/reviews/openslide/openslide/157#-KT7c7mnVBf-90Mbq-Fg:-KT7c7mnVBf-90Mbq-Fh:biw3qxn) ([raw file](https://github.com/openslide/openslide/blob/ac517ffa5030e851d4373e082e4a7f1bc3d37a7f/src/openslide-decode-dicom.c#L956)):*
> ```C
>     add_tag_path( instance->ds.tps, tp );
>     destroy_path( tp );
>     }
> ```

As I understand it, a DICOMDIR can describe objects from multiple studies or series, and a whole-slide image corresponds to a single series.  If that is correct, we should detect and reject DICOMDIRs referring to multiple studies or series, since our API provides no way for the user to select the one they want.

---

*[src/openslide-vendor-dicom.c, line 168 at r1](https://reviewable.io:443/reviews/openslide/openslide/157#-KT78jnbIA4BadslELTj:-KT78jnbIA4BadslELTk:b9k8cm5) ([raw file](https://github.com/openslide/openslide/blob/ac517ffa5030e851d4373e082e4a7f1bc3d37a7f/src/openslide-vendor-dicom.c#L168)):*
> ```C
>   }
> 
>   return true;
> ```

We need to reject non-WSMIS DICOM files so the application will know to try other decoders (such as ImageMagick) which might support the file.

---

*[src/openslide-vendor-dicom.c, line 217 at r1](https://reviewable.io:443/reviews/openslide/openslide/157#-KT7YzB5sWUIjW7TsxIx:-KT7YzB5sWUIjW7TsxIy:b-1312l7) ([raw file](https://github.com/openslide/openslide/blob/ac517ffa5030e851d4373e082e4a7f1bc3d37a7f/src/openslide-vendor-dicom.c#L217)):*
> ```C
>     struct _openslide_dicom * instance = _openslide_dicom_create(*fullpath, err);
>  
>     if (!_openslide_dicom_level_init(instance,
> ```

This assumes a 1:1 mapping between levels and image objects, but that's not necessarily true:
- Multiple Z-planes in a level can be stored in different objects (Sup. 145 lines 293-294).
- Likewise with multiple wavelengths in fluorescence imaging (lines 294-295).
- There can be multiple tile sizes and grids in a level (lines 303-306).

---

*[src/openslide-vendor-dicom.c, line 223 at r1](https://reviewable.io:443/reviews/openslide/openslide/157#-KT7mmBiPJ-fbUUNFdl7:-KT7mmBiPJ-fbUUNFdl8:b-cdi1hy) ([raw file](https://github.com/openslide/openslide/blob/ac517ffa5030e851d4373e082e4a7f1bc3d37a7f/src/openslide-vendor-dicom.c#L223)):*
> ```C
>       g_slice_free(struct level, l);
>     }
>     _openslide_dicom_destroy(instance);
> ```

We need to reject the slide unless:
- images are visible-light rather than fluorescence,
- there is only one Z-plane, and
- tiles are JPEG-encoded.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openslide/openslide/157)*

bgilbert:
*[src/openslide-decode-dicom.c, line 74 at r1](https://reviewable.io:443/reviews/openslide/openslide/157#-KQJVHommHOnxkjNVQUd:-KT89iIVc0j-CgfmXggJ:by88eh6) ([raw file](https://github.com/openslide/openslide/blob/ac517ffa5030e851d4373e082e4a7f1bc3d37a7f/src/openslide-decode-dicom.c#L74)):*
<details><summary><i>Previously, bgilbert (Benjamin Gilbert) wrote…</i></summary>
> Is this a scenario likely to be encountered by users?  If so, we should be able to read the files as-is, without requiring an external indexing step.
</details>
It seems the DICOMDIR is necessary when there are multiple DICOM files, but we could still avoid requiring it when reading a single file.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openslide/openslide/157)*

bgilbert:
Review status: all files reviewed at latest revision, 40 unresolved discussions.

---

*[a discussion](https://reviewable.io:443/reviews/openslide/openslide/157#-KQJRDxZIqlUwwwmpxY8:-KTVzsOzLi6L65A4AqNt:bvuhwmg) (no related file):*
<details><summary><i>Previously, bgilbert (Benjamin Gilbert) wrote…</i></summary>
> I think the major question is whether we should really be implementing our own DICOM parser, or whether we should use an existing DICOM library instead.  Could you comment on the tradeoffs from your perspective?  The decoder implemented here is already a lot of code, and apparently doesn't handle cases (missing DICOMDIR) that we may need to support.  I'd rather avoid maintaining this complexity in OpenSlide if possible, especially if it requires duplicating code already available elsewhere.  (The only reason we implement our own TIFF parser is that libtiff can't handle NDPI.)
> 
> As I understand it, we could use a C++ DICOM library by implementing openslide-decode-dicom in C++, exposing only C interfaces.  We'd need to link OpenSlide using the C++ linker, but as far as I can tell, users of the library wouldn't see any difference.
</details>
Resolved in favor of a custom parser ([thread](https://lists.andrew.cmu.edu/pipermail/openslide-users/2016-October/001317.html)).

---

*[src/openslide-decode-dicom.c, line 56 at r1](https://reviewable.io:443/reviews/openslide/openslide/157#-KQJe84uyuc1NYkrWToO:-KTW-34nAyiCYdt1L_Hb:b3bhqpa) ([raw file](https://github.com/openslide/openslide/blob/ac517ffa5030e851d4373e082e4a7f1bc3d37a7f/src/openslide-decode-dicom.c#L56)):*
<details><summary><i>Previously, bgilbert (Benjamin Gilbert) wrote…</i></summary>
> This comment should go in the vendor driver, not here.
> 
> See quickhash discussion [here](https://lists.andrew.cmu.edu/pipermail/openslide-users/2015-March/001030.html).  Also, we'll want to hash the lowest-resolution pyramid level in case of UID collisions.
</details>
Now that I understand this better:

- AIUI, if there are multiple WSI series in a study, those series represent completely independent WSI with different parameters.  If so, the Series Instance UID should certainly be hashed into the quickhash.
- I now think we shouldn't hash the lowest-resolution level.  The WG26 Hamamatsu samples have smallest levels which are still quite large, and that has [caused performance problems before](https://github.com/openslide/openslide/issues/79).  However, I'd still prefer to have some insurance against UID collisions.  Perhaps we could hash in the acquisition datetime, image dimensions, and whatever other attributes seem useful.
- It's clearly possible to generate a DICOMDIR for individual pieces of an acquisition; are we likely to encounter this?  The quickhash of a complete series (including all pyramid levels, label image, etc.) should be different than (for example) the quickhash of a single pyramid level from that series.  Is it reasonable to handle that case by hashing in a sorted list of unique identifiers for each object in the series?

---

*[src/openslide-decode-dicom.c, line 956 at r1](https://reviewable.io:443/reviews/openslide/openslide/157#-KT7c7mnVBf-90Mbq-Fg:-KTW4z3r9kZTHhFdJv5y:b-ro5l90) ([raw file](https://github.com/openslide/openslide/blob/ac517ffa5030e851d4373e082e4a7f1bc3d37a7f/src/openslide-decode-dicom.c#L956)):*
<details><summary><i>Previously, bgilbert (Benjamin Gilbert) wrote…</i></summary>
> As I understand it, a DICOMDIR can describe objects from multiple studies or series, and a whole-slide image corresponds to a single series.  If that is correct, we should detect and reject DICOMDIRs referring to multiple studies or series, since our API provides no way for the user to select the one they want.
</details>
DICOM is the only supported slide format where one filename can refer to multiple independent WSIs.  I'm hesitant to handle that by adding virtual filesystem functions to the API, since that would duplicate functionality that should be handled by the actual filesystem.  (And, in practice, applications might ignore the new API.)  As a workaround, the DICOM driver may have to support fake directory components, e.g. `/path/to/DICOMDIR/<series UID>`.

Better ideas?

---

*[src/openslide-vendor-dicom.c, line 223 at r1](https://reviewable.io:443/reviews/openslide/openslide/157#-KT7mmBiPJ-fbUUNFdl7:-KTW5xqIiJBd85jSbk9Y:bpe3h4k) ([raw file](https://github.com/openslide/openslide/blob/ac517ffa5030e851d4373e082e4a7f1bc3d37a7f/src/openslide-vendor-dicom.c#L223)):*
<details><summary><i>Previously, bgilbert (Benjamin Gilbert) wrote…</i></summary>
> We need to reject the slide unless:
> - images are visible-light rather than fluorescence,
> - there is only one Z-plane, and
> - tiles are JPEG-encoded.
</details>
I was wrong about Z-planes.  Other drivers currently select a default Z-plane (typically the center one) and ignore the others.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openslide/openslide/157)*

malaterre:
*[a discussion](https://reviewable.io:443/reviews/openslide/openslide/157#-KQJRg2nXIzeHG8GXx99:-Kerhkgu0TBZe3ozr-Bp:b6ijlln) (no related file):*
<details><summary><i>Previously, bgilbert (Benjamin Gilbert) wrote…</i></summary><blockquote>

The test suite will need test cases for this format, which means we'll need some redistributable sample data.  I'm not aware of any public sample files which are explicitly redistributable; can you provide some?
</blockquote></details>

A DICOM Series is available at ftp://medical.nema.org/MEDICAL/Dicom/DataSets/WG26/Hamamatsu/ I need to double check but the license to copy & host them should be pretty flexible. What kind of permission do you want to have ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openslide/openslide/157)*

malaterre:
*[a discussion](https://reviewable.io:443/reviews/openslide/openslide/157#-KQJlIphM_ULmD62ZYEA:-KeriClg5pFWCCensGWd:b-h3zsia) (no related file):*
<details><summary><i>Previously, bgilbert (Benjamin Gilbert) wrote…</i></summary><blockquote>

There is no support for properties and associated images.  This should be added, though perhaps not immediately.
</blockquote></details>

What would be the minimal list of properties that you would like to see ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openslide/openslide/157)*

malaterre:
assert() violation have been fixed https://github.com/malaterre/openslide/commit/223d3b177e83e18aa10177b834019810a23d196c

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/openslide/openslide/157#-:-KeryC7oYFDPdZorTvVM:bqxsndn)*

malaterre:
As side note I know some compiler are using `assert` to guide the optimizer so it should be made clear the OpenSlide coding style is discarding this particular kind of vendor specific optimization.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/openslide/openslide/157#-:-KerywnvdA6pQguMJSD8:b-azjcy3)*

malaterre:
*[src/openslide-decode-dicom.c, line 816 at r1](https://reviewable.io:443/reviews/openslide/openslide/157#-KQJZIBovaSFnZsJpgow:-Ket-aNcmRz0vDept3tq:b-uvt4iy) ([raw file](https://github.com/openslide/openslide/blob/ac517ffa5030e851d4373e082e4a7f1bc3d37a7f/src/openslide-decode-dicom.c#L816)):*
<details><summary><i>Previously, bgilbert (Benjamin Gilbert) wrote…</i></summary><blockquote>

glib has a function for this.
</blockquote></details>

Should be fixed in latest commit

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openslide/openslide/157)*

bgilbert:
We're not avoiding assertions entirely; we just don't want to crash the whole program on bad input data.  You're correct that this should be documented; I've filed [a bug](https://github.com/openslide/openslide.github.com/issues/10).

Thanks for your patience with this PR.  I'll get back to it when I can.

---

Review status: 2 of 6 files reviewed at latest revision, 40 unresolved discussions.

---

*[a discussion](https://reviewable.io:443/reviews/openslide/openslide/157#-KQJRg2nXIzeHG8GXx99:-Kkt7ygiKiFeKj1H1Q1h:b-j55sem) (no related file):*
<details><summary><i>Previously, malaterre (Mathieu Malaterre) wrote…</i></summary><blockquote>

A DICOM Series is available at ftp://medical.nema.org/MEDICAL/Dicom/DataSets/WG26/Hamamatsu/ I need to double check but the license to copy & host them should be pretty flexible. What kind of permission do you want to have ?
</blockquote></details>

Ideally, Creative Commons Zero.  In practice, any broadly permissive license should work.  We do need to be able to modify the samples (for test cases), and probably to distribute modified copies.  As far as I can tell, though, there's currently no published license at all.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openslide/openslide/157#-:-Kkt785O_d-LpSch9iH1:bclh3jt)*

feliwir:
Can this please be merged? What's missing here, i'd happily do the required changed, since this is a requirement for me.

feliwir:
any news @malaterre ?

malaterre:
@feliwir Sorry I've lost interest in this PR. The code should be relatively easy to read, and refactor to meet the coding-style/requirements needed for OpenSlide. Feel free to hack. Good luck !

feliwir:
@malaterre ah that's bad news, because i have a DICOM that doesn't work with your branch and i am not really sure what's the cause of that. Do you support pathological dicom images in GDCM? (>1GB)

malaterre:
@feliwir What do you mean by "does not work" ? Does it crash the DICOM parser ? Trigger an assert() ? Or rendering is invalid...

feliwir:
@malaterre i couldn't quite figure out what was wrong. The DICOM file in question is [this one](https://mbitsimaginggmbh-my.sharepoint.com/:u:/g/personal/vedder_mbits_info/ERggZYnrPY5Fv9TBrrVMe-8BuX_vCUuLtb9jlH0EdEARPw?e=29F8qu).

It was definetly picked up correctly by your DICOM backend, but then failed some check, which i couldn't quite figure out.

feliwir:
Did you have a chance to take a look @malaterre ?

bgilbert:
Closing in favor of #431.  I really appreciate the effort you put in here, even if the code didn't end up landing.

Issue #171, opened 2016-01-22
Hamamatsu integer overflow when parsing VMS


cryptoad:
num_cols & num_rows are integers read from the VMS.
They are multiplied to compute num_images, also an integer, here:
https://github.com/openslide/openslide/blob/master/src/openslide-vendor-hamamatsu.c#L1853

The multiplication can overflow, ie: if num_cols is 65536 and num_rows is 65536, num_images will be 1.

This will later allow the image_filenames array to be referenced out of bounds.
The checks here:
https://github.com/openslide/openslide/blob/master/src/openslide-vendor-hamamatsu.c#L1952
will pass, and this can end up in memory corruption here:
https://github.com/openslide/openslide/blob/master/src/openslide-vendor-hamamatsu.c#L1972

An example VMS triggering the issue:
```
[Virtual Microscope Specimen]                                                   
NoLayers=1                                                                      
NoJpegColumns=65536                                                             
NoJpegRows=65536                                                                
ImageFile(100,0)=CMU-1-40x - 2010-01-12 13.24.05.jpg                              
MapFile=CMU-1-40x - 2010-01-12 13.24.05_map2.jpg                                
OptimisationFile=CMU-1-40x - 2010-01-12 13.24.05.opt                            
AuthCode=635108335                                                              
SourceLens=40.000000                                                            
PhysicalWidth=23367500                                                          
PhysicalHeight=17357904                                                         
LayerSpacing=2032236150                                                         
MacroImage=CMU-1-40x - 2010-01-12 13.24.05_macro.jpg                            
PhysicalMacroWidth=76000000                                                     
PhysicalMacroHeight=26000000                                                    
XOffsetFromSlideCentre=4876667                                                  
YOffsetFromSlideCentre=-2340000                                                 
Reference=CMU-1-40x                                                             
```

bgilbert:
Thanks for the report!

PR #172, opened 2016-01-23
Fix an integer overflow in Hamamatsu parsing

agoode:
We can overflow when reading a VMS file. This is bad since we index on the value later and can do an out of bounds memory access.
Thanks to Kostya Kortchinsky.

Addresses issue #171.


bgilbert:
Thanks for the fix!

Issue #174, opened 2016-02-04
Certain (very large) Hamamatsu NDPI files cannot be opened with OpenSlide


GeertLitjens:
I have some Hamamatsu slides which are very large (>5GB) which cannot be opened (error message is "Can't validate JPEG for directory 0: Expected marker at 4294971018, found none". I am able to share such a file if this is needed. They do open with Bioformats and NDPIView.

ilykos:
Same issue here.
This concerns the 50 x 75 slide format.

ylaizet:
Hello, I used the openslide-python lib and got the same issue on Hamamatsu file > 4.2Go. None of the big files can be opened with openslide while the windows ndpi viewer works perfectly on those big files.
Is there a workaround?

CamNZ:
Are you trying to load the entire slide into memory? This is a common source of errors with large WSI files (instead just load parts of the slide as you need them)

bgilbert:
@CamNZ That is indeed a common source of errors, but this is a real bug.

@ylaizet The workaround, for what it's worth, is to keep pyramid levels smaller than 4 GiB (or perhaps 2 GiB) -- by reducing either the scan resolution or the size of the scan area.

All: this bug is on my radar, but I may not have a chance to get to it for a while.  It's actually more of a missing feature, and the fix may not be trivial.

ylaizet:
Hey, I use the openslide-python [deepzoom_multiserver.py](https://github.com/openslide/openslide-python/blob/master/examples/deepzoom/deepzoom_multiserver.py) script to set up a server which lists all available slides by using the lowlevel.detect_vendor function - which maps to the [openslide_detect_vendor](https://github.com/openslide/openslide-python/blob/master/openslide/lowlevel.py) function - to detect the files. As far as I could see, from there, I get None (NULL) from the detection on big files whereas smaller files return "hamamastu". I guess, this does not load the entire file yet. Actually, the script runs on a 32Go memory computer so I would be surprised to have a lack of memory but there might be a parameter to adjust?
I keep in mind that for the moment, it would be better to try to have small files. I am looking forward to be able to load bigger file. If you have any idea to make it work in the meantime, I would be happy to give it a try.
Thanks.

ilykos:
> Is there a workaround?

Hi @ylaizet, I've had the same problem and one of the alternatives I considered is to use bioformats python library as @GeertLitjens suggest. Though, I have not tried it myself. I can also mention that this is not the most elegant solution since it depends on java.

[...]

PR #211, opened 2017-07-24
Fix to read Aperio svs images with one or more "broken" tiles

hamohn:
In some Aperio images, one tile (typically 0, 0), in one or more levels, is partly overwritten with some unknown data. This seems to happen from byte 0 or byte 1 of the tile data, thereby overwriting the JPEG marker. The first erroneous byte always seems to be a 0x11, so it is very easy and safe to detect the bug.

In such cases Openslide would just return a black image.

The fix is to return false from _openslide_tiff_read_tile() when the erroneous data is detected, and to call render_missing_tile() if _openslide_tiff_read_tile() retruns false.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,7 @@ Makefile.in
 /stamp-h1
 /*.pc
 
+.vscode
 /mingw32-config.cache
 
 /src/make-tables
--- a/src/openslide-decode-tiff.c
+++ b/src/openslide-decode-tiff.c
@@ -299,6 +299,17 @@ bool _openslide_tiff_read_tile(struct _openslide_tiff_level *tiffl,
       return false;
     }
 
+    // In some Aperio images, one tile (typically 0, 0), in one or more levels,
+    // is partly overwritten with some unknown data. This seems to happen
+    // from byte 0 or byte 1 of the tile data. Such garbled tiles are
+    // impossible to read, so just return false!
+    // The other half of this fix/hack consists of calling render_missing_tile
+    // from decode_tile (in openslide-vendor-aperio) if _openslide_tiff_read_tile
+    // returns false.
+    const unsigned char *bytes = buf;
+    if (bytes[0] == 0x11 || (bytes[0] == 0xff && bytes[1] == 0x11))
+      return false; // Garbled tile
+
     // decompress
     bool ret = decode_jpeg(buf, buflen, tables, tables_len,
                            tiffl->photometric == PHOTOMETRIC_YCBCR ? JCS_YCbCr : JCS_RGB,
--- a/src/openslide-vendor-aperio.c
+++ b/src/openslide-vendor-aperio.c
@@ -159,12 +159,11 @@ static bool decode_tile(struct level *l,
     break;
   default:
     // not for us? fallback
-    bool ok = _openslide_tiff_read_tile(tiffl, tiff, dest,
-                                     tile_col, tile_row,
-                                     err);
     // If the tile could not be read/rendered, (maybe because of garbled tile data)
     // try to render it using data at other levels using render_missing_tile()
-    return ok || render_missing_tile(l, tiff, dest, tile_col, tile_row, err);
+    return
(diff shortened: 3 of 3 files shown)


.gitignore
```diff
@@ -33,6 +33,7 @@ Makefile.in
 /stamp-h1
 /*.pc
 
+.vscode
```
bgilbert: We don't include editor-specific configuration in `.gitignore`; please remove.  You can use the `core.excludesFile` Git setting to configure this globally on your system.

src/openslide-decode-tiff.c
```diff
@@ -299,6 +299,17 @@ bool _openslide_tiff_read_tile(struct _openslide_tiff_level *tiffl,
...
 
+    // In some Aperio images, one tile (typically 0, 0), in one or more levels,
+    // is partly overwritten with some unknown data. This seems to happen
+    // from byte 0 or byte 1 of the tile data. Such garbled tiles are
+    // impossible to read, so just return false!
+    // The other half of this fix/hack consists of calling render_missing_tile
+    // from decode_tile (in openslide-vendor-aperio) if _openslide_tiff_read_tile
+    // returns false.
+    const unsigned char *bytes = buf;
+    if (bytes[0] == 0x11 || (bytes[0] == 0xff && bytes[1] == 0x11))
```
bgilbert: This shouldn't be in driver-independent code.  Instead, I'd suggest that the Aperio `decode_tile` handle the error from `_openslide_tiff_read_tile` by calling `_openslide_tiff_read_tile_data` and checking for the bad data itself.  Since this will rarely happen, the performance penalty should be acceptable.

hamohn:
Hi *@bgilbert*
This is my first attempt at contributing to an open source project, so
please bear with me :-)
Thank you for insightful comments, sounds like good points! The first one
is easy to deal with obviously, but I'm struggling with the second!
The problem is that if we let _openslide_tiff_read_tile call decode_jpeg,
it bails out in "a big way" by calling longjmp or something like that. It
never returns!
And it's pretty hard to debug this on Windows - the only technique I have
found is good old printf-debugging ;-)

------------------------------

Hans Martin Mohn |  Gjørstadsgate 4, 0367 Oslo |  93 22 94 30 |
hans.martin.mohn@gmail.com

On Wed, Jul 26, 2017 at 10:01 AM, Benjamin Gilbert <notifications@github.com

bgilbert:
No problem. Thanks for contributing! 😃

The longjmp is supposed to happen inside `decode_jpeg`, and then `decode_jpeg` is supposed to return normally.  What happens instead?  Does the program crash?

I haven't found a great way to debug on Windows either.  gdb inside Cygwin mostly works, if the calling program is built with MinGW.  If you'd like to try that, you can use `openslide-write-png.exe` for testing.  It'd probably be a good idea to test with `openslide-write-png` anyway, since the longjmp problem may only happen when called from MSVC-compiled code.

bgilbert:
https://github.com/openslide/openslide/issues/174#issuecomment-316412344 sounds like the same problem, and reportedly occurs with slides larger than ~4 GB.  How large are the affected slide files you're seeing?

hamohn:
It's the same actual problem - we work in the same company :-)

------------------------------

Hans Martin Mohn |  Gjørstadsgate 4, 0367 Oslo |  93 22 94 30 |
hans.martin.mohn@gmail.com

On Sat, Jul 29, 2017 at 10:53 AM, Benjamin Gilbert <notifications@github.com

bgilbert:
Hi @hamohn, have you had a chance to look at this further?

lferreiraMD:
@hamohn Hi Hans,
So, basically, this hasn't been solved yet? I've just got this error when trying to open a +4GB Aperio `.svs` file.
Cheers,
Leo

hamohn:
No, sorry. We are just using this fork, I don't have time do it properly.

lør. 8. mai 2021, 21:05 skrev Leonardo Ferreira ***@***.***>:

PR #236, opened 2018-03-23
fix crash on Alpine Linux with glib 2.50+ --- remove explicit g_type_init()

jcupitt:
`g_type_init()` has not been needed since glib 2.36. Calling it anyway is
normally harmless, except on Alpine Linux with glib 2.50+, where it
triggers an assert fail due to differences in the way the libcs handle
startup.

This patch stop openslide calling `g_type_init()` if glibc is later than
2.36.

Related issue: https://github.com/jcupitt/libvips/issues/874

### DIFF
--- a/src/openslide.c
+++ b/src/openslide.c
@@ -60,8 +60,10 @@ static void __attribute__((constructor)) _openslide_init(void) {
   if (!g_thread_supported()) {
     g_thread_init(NULL);
   }
-  // initialize GObject
+  // not needed since glib 2.36, and causes problems with musl on 2.50+
+#if !GLIB_CHECK_VERSION(2, 36, 0)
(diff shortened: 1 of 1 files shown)


bgilbert:
@jcupitt Thanks for the PR, and apologies for the three-year delay.  #326 drops support for older glib versions and thus drops the call entirely.  I'll close this PR in favor of that one.

jcupitt:
No problem, I'm glad you're working through the backlog! Does this mean 3.5 might be coming soon?

bgilbert:
I'm doing some infrastructure cleanup, but I'm not going to be able to work through the entire backlog unfortunately.  We should probably do a 3.4.2 with accumulated fixes though.

Issue #234, opened 2018-03-12
Ventana BIF file : Bad direction attribute "LEFT"


ylaizet:
## Context

**Issue type** (bug report or feature request): bug report
**Operating system** (e.g. Fedora 24, Mac OS 10.11, Windows 10): Ubuntu 16.04
**Platform** (e.g. 64-bit x86, 32-bit ARM): 64-bit x86
**OpenSlide Python version** (`openslide.__version__`): 1.1.1
**OpenSlide version** (`openslide.__library_version__`): 3.4.0-1build1
**Slide format** (e.g. SVS, NDPI, MRXS): bif

## Details

I use openslide-python successfully to view ndpi files, then I tried to view .bif slides but I get an error:

`Bad direction attribute "LEFT"`

I can open my bif file with a slide viewer but not with openslide. If I use the example OS-2bif file, openslide works so I guess that my bif file may have special attribute specification not supported in openslide.

## EDIT:

I tried to dig into the error and I came up with the following:
The error is raised in the [openslide-vendor-ventana.c file line 589](https://github.com/openslide/openslide/blob/7b99a8604f38280d14a34db6bda7a916563f96e1/src/openslide-vendor-ventana.c#L590) where the value of the direction variable is obviously set to "LEFT" in my .bif file. It seems that this value is not supported (the reason in unknown to me, wether my file is no properly formatted or of MEFT is a missing value here?).
At the top of the file, I see only "RIGHT" and "UP" constant values declared but no "LEFT", which ends to the encountered error. Does anybody have enough experience with the .bif format specification or openslide lib code to give an hint to fix this?

Thanks

PS : Sorry for duplicating the post, I first posted on the python interface project and realized that this is rather related to the C library.

sbengali:
I'm using OpenSlide on Python and I'm running into the same error when trying to read a BIF file. I have the latest version so there's nothing to upgrade. Some help would be much appreciated.

kjmin-deepbio:
The "openslide-show-properties file.bif" command also produces the same error.

Rasaa84:
Is this problem not solved yet in new version of openslide?

Himanshunitrr:
Will be helpful, if its extended to these formats too

Himanshunitrr:
the workaround it is to convert those .bif files to .tif files using [tifffile](https://github.com/DigitalSlideArchive/tifftools)

killerz99:
import glob
import tifftools

# read all .bif files in the current directory
[...]

PR #246, opened 2018-07-23
Support for Tile Joint Left Direction

ChrisSmith96:
I have access to some .bif files which have the TileJointInfo direction attribute set to LEFT rather than RIGHT. Openslide doesn’t currently support such files so I have made some simple changes so they can be opened. I have done some testing opening .bif files containing the LEFT join direction and the resulting image is almost exactly the same when processed as if it had the direction attribute set to RIGHT. Please could you consider merging this change into openslide to support files with this attribute!

### DIFF
--- a/src/openslide-vendor-ventana.c
+++ b/src/openslide-vendor-ventana.c
@@ -67,6 +67,7 @@ static const char ATTR_TILE1[] = "Tile1";
 static const char ATTR_TILE2[] = "Tile2";
 static const char ATTR_OVERLAP_X[] = "OverlapX";
 static const char ATTR_OVERLAP_Y[] = "OverlapY";
+static const char DIRECTION_LEFT[] = "LEFT";
 static const char DIRECTION_RIGHT[] = "RIGHT";
 static const char DIRECTION_UP[] = "UP";
 
@@ -572,7 +573,7 @@ static struct bif *parse_level0_xml(const char *xml,
       bool ok;
       bool direction_y = false;
       //g_debug("%s, tile1 %"PRId64" %"PRId64", tile2 %"PRId64" %"PRId64, (char *) direction, tile1_col, tile1_row, tile2_col, tile2_row);
-      if (!xmlStrcmp(direction, BAD_CAST DIRECTION_RIGHT)) {
+      if (!xmlStrcmp(direction, BAD_CAST DIRECTION_RIGHT) || !xmlStrcmp(direction, BAD_CAST DIRECTION_LEFT)) {
         // get left joint of right tile
         struct tile *tile =
           area->tiles[tile2_row * area->tiles_across + tile2_col];


agoode:
Thanks for this contribution.

I want to make sure we are supporting this variant properly.
* What are the differences between `LEFT` and `RIGHT`?
* Should we add an OpenSlide property to expose this?

ChrisSmith96:
I’m not sure that I understand the differences between the two values, but all images that I have access to from a VENTANA DP 200 scanner have this attribute set to LEFT and not to RIGHT. 

With regards to adding an OpenSlide property, I don’t think that this is a generic property seen in all files. It appears to be specific to BIF files.

jrshoare:
This PR looks as though it might solve issue #234. Any idea when it might get merged in to the base branch?

p829911:
I had the same problem with the ventana`.bif` file, so I compiled myself and used openslide.
May I ask you to merge this pull request?

jimtalksdata:
Agree on merging this pull request. Only changes the appropriate file and one conditional, so I don't see any backwards or forwards compatibility issues.

sbengali:
Is there any update to the python package that fixes this issue? I am trying to read BIF files and running into this problem. Thanks.

suk-kim:
Is this problem not solved yet?

The left problem has been solved, but the problem of the image not opening properly is not solved.

ValentinFrancois:
Hi, would this PDF help?
There's a whole section about the image stitching process.  
[Roche Digital Pathology - _BIF image file format for digital pathology_](https://diagnostics.roche.com/content/dam/diagnostics/Blueprint/en/pdf/rmd/Roche-Digital-Pathology-BIF-Whitepaper.pdf)

bgilbert:
@ValentinFrancois The BIF specification that you referenced appears to be covered by [these license terms](https://diagnostics.roche.com/global/en/legal/bif-specifications-license-terms-n-conditions.html), which right away in the first paragraph contain a field-of-use restriction (the definition of "Purpose" and the limitation to certain scanner models) which is incompatible with OpenSlide's LGPL license.  As a result, we cannot accept any code into OpenSlide that has been created with the aid of that specification document.  I've marked your comment as "off-topic" so GitHub's web interface will hide it by default.

ValentinFrancois:
@bgilbert Sorry for that, the PDF is referenced in Google so one can access it outside of the download form without being aware of the license. Thanks for your alertness.

bgilbert:
It appears we'll need more than this small change to support these slides correctly.  When opening [Ventana-1.bif](https://openslide.cs.cmu.edu/download/openslide-testdata/Ventana/Ventana-1.bif) with `LEFT` permitted, the image is offset to the right by roughly the width of the image, suggesting that the X origin is on the right-hand side.  (This sort of thing is the reason OpenSlide refuses to open slides with unexpected metadata.)

I'll go ahead and close this.  Thanks for the PR.

PR #276, opened 2019-07-29
Add support for NDPI images >4GB

ntrahearn:
Addresses issue #174 (and possibly others), regarding the reading of large NDPI files. This issue stems from the fact that classical TIFF (and by extension, NDPI) images only support up to 32 bit values for tagged metadata. However, due to the nature of whole slide image data, it isn't uncommon for the size of an NDPI to exceed the 32 bit range. This means that key metadata, such as the byte positions of image layers or their size in bytes, may be too large to store in a traditional TIFF IFD entry.

Currently OpenSlide relies on heuristics to determine the high bits of 64 bit addresses, which fail in some cases. However, this is unnecessary, as NDPI actually stores the high bits of the offset/value of each tag in 4 byte blocks immediately after the end of the IFD.

This fix modifies openslide-decode-tifflike.c to append these extra 4 bytes to each IFD entry's value/offset and, if necessary, modifies its type to LONG8.

This fix also modifies openslide-vendor-hamamatsu.c to construct correct restart marker addresses. Currently only the values in TIFF tag 65426 are used, which are only the lower 32 bits of each address. High bits are stored in TIFF tag 65432, so these are now appended before mcu_starts are calculated

### DIFF
--- a/src/openslide-decode-tifflike.c
+++ b/src/openslide-decode-tifflike.c
@@ -175,18 +175,6 @@ static uint32_t get_value_size(uint16_t type, uint64_t *count) {
   }
 }
 
-// Re-add implied high-order bits to a 32-bit offset.
-// Heuristic: maximize high-order bits while keeping the offset below diroff.
-static uint64_t fix_offset_ndpi(uint64_t diroff, uint64_t offset) {
-  uint64_t result = (diroff & ~(uint64_t) UINT32_MAX) | (offset & UINT32_MAX);
-  if (result >= diroff) {
-    // ensure result doesn't wrap around
-    result = MIN(result - UINT32_MAX - 1, result);
-  }
-  //g_debug("diroff %"PRIx64": %"PRIx64" -> %"PRIx64, diroff, offset, result);
-  return result;
-}
-
 #define ALLOC_VALUES_OR_FAIL(OUT, TYPE, COUNT) do {			\
     OUT = g_try_new(TYPE, COUNT);					\
     if (!OUT) {								\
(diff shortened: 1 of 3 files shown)


andor-pierdelacabeza:
Interesting!

AndyBlake:
hiya - tried this with a 30gb ndpi and cannot get it work - same issue arises:

openslide-write-png: ./largeslide.ndpi: Can't validate JPEG for directory 0: Expected marker at 21474841381, found none
 for example. 
this is after compiling from source from your master branch https://github.com/ntrahearn/openslide

andor-pierdelacabeza:
That's a big file, @AndyBlake . What's the resolution for that? What device did you use for scanning? (Just out of curiosity)

JiweiMaster:
@ntrahearn Hi, ntrahearn, Do you know how to use this PR on windows?, maybe you have windows version openslide of this PR. Look forward to your reply，thanks.

Signynt:
Works great on Ubuntu 22.04 LTS, makes working with files >4GB the same as smaller ones.

bgilbert:
@ntrahearn Thanks very much for tracking this down!  Would you be willing to rebase your changes onto current Git `main`?  There have been several internal API changes since you originally did this work.

Also, please [certify to the Developer Certificate of Origin](https://openslide.org/docs/signoff/) by adding signoffs to your commit messages.  Even if you don't have time to rebase your changes onto `main`, if you can update your commit messages in-place to add signoffs, I can handle updating the PR and getting it landed.

ntrahearn:
Sure, I've made the changes and opened a fresh pull request here:

https://github.com/openslide/openslide/pull/602

bgilbert:
Thanks!  Closing in favor of your updated PR #602.

Issue #251, opened 2018-09-26
 Mrxs Error: Cannot read slide position info: Expected 1 value


jetic83:
## Context

**Issue type** (bug report or feature request): bug report
**Operating system** (e.g. Fedora 24, Mac OS 10.11, Windows 10): Win 7
**Platform** (e.g. 64-bit x86, 32-bit ARM): 64-bit
**OpenSlide version**: 3.4.1
**Slide format** (e.g. SVS, NDPI, MRXS): MRXS

## Details
When using openslide-show-properties.exe, I get an error with certain (not all) .mrxs files:

`Cannot read slide position info: Expected 1 value`

However, ImageScope and CaseViewer can open those files without problems.

0x001B:
## Context

**Operating system** (e.g. Fedora 24, Mac OS 10.11, Windows 10): RHEL 7
**Platform** (e.g. 64-bit x86, 32-bit ARM): 64-bit
**OpenSlide version**: 3.4.1
**Slide format** (e.g. SVS, NDPI, MRXS): MRXS

## Details

I get exactly the same error with Openslide 3.4.1 on a RHEL 7 system for certain (not all) .mrxs files.

Any news on this issue?

Evizero:
I also see the same error for a couple of slides

cett01:
I saw the same error for a couple of slides, on  version 3.4.1. CaseViewer can open this slides and zoom perfectly to level 0. I can share this slides with you.
`java.io.IOException: Cannot read slide position info: Expected 1 value
	at org.openslide.OpenSlide.checkError(OpenSlide.java:178)
	at org.openslide.OpenSlide.<init>(OpenSlide.java:114)
`

mjkvaak:
I was able to solve this problem with Slide Converter provided by 3D Histec in their webpage: https://www.3dhistech.com/software-downloads. Simply convert from mrxs to mrxs after which openslide can open the images. I will report, if I can find some differences between the original and the converted files.

skrackow:
The difference is that the conversion removes the overlap of camera images.

iowatom:
I am having the same issue with a very few mrxs file having the "Expected 1 value" when I try to extract to a .dz file. 
I am not understanding how or whether this can be solved from these comments. Any clarification would be appreciated.

iowatom:
[...]

PR #293, opened 2020-05-11
support for some 3dhistech scanners

koltenpearson:
#251


maksimkurb:
Confirm that it works for my slides

src/openslide-vendor-aperio.c
```diff
@@ -558,6 +572,12 @@ static bool aperio_open(openslide_t *osr,
     } else {
       // associated image
       const char *name = (dir == 1) ? "thumbnail" : NULL;
+      if (gt450_workaround) {
+          if(dir == tiff_dir_count-2)
+              name = "label";
+          if (dir == tiff_dir_count-1)
+              name = "macro";
```
fiendish, line 579: There's a comment in https://github.com/openslide/openslide/issues/297 that you should check TIFFTAG_SUBFILETYPE instead of doing this. This will mislabel frames in files that don't have a label or macro frame (they're optional after all).

PR #311, opened 2020-09-30
Update README.txt

abhigyan1000:


jaharkes:
I don't see how this improves on what is already there.

PR #312, opened 2020-09-30
Update spell Docs README.txt

mrvlsaf:
Spell Docs Updated.


jaharkes:
The original text is actually correct, it parses as "this library reads (whole slide images)" while your correction changes the meaning to "this library (reads whole) slide images".

Also it really should be "this library reads", and not "this library read".

PR #326, opened 2021-04-08
Drop support for libtiff 3, OpenJPEG 1.x, and glib < 2.56

bgilbert:
libtiff 4.0.0 is more than nine years old and OpenJPEG 2.1.0 is seven years old.  glib &ge; 2.56 is available in Debian stable, all non-EOL Ubuntu releases except 16.04 (which is reaching "end of standard support" this month), and RHEL 7 and 8.


test/cases/mirax-slidedat-bad-image-concat-factor/config.yaml
```diff
@@ -1,5 +1,5 @@
 base: Mirax/CMU-1.zip
-error: ^Key file contains key ('|“)IMAGE_CONCAT_FACTOR('|”) in group ('|“)LAYER_0_LEVEL_2_SECTION('|”) which has (a )?value that cannot be interpreted.$
```
agoode, line 2: How is this change related to the new glib version?

agoode (approved):

test/cases/mirax-slidedat-bad-image-concat-factor/config.yaml
bgilbert: The regex matches a glib error string.  We can stop matching variants produced by older glib versions.

PR #329, opened 2021-05-24
Integrate support for virtual file systems

### LINKED ISSUE #200: Openslide can read the remote slide on internet?
Hi All, 
I have a project need to use the Openslide to read the remote file,as like "http://www.xxx.com/xxx.ndpi"
is it working? Our development language is C#

## Context

**Issue type** (bug report or feature request): 
**Operating system** (e.g. Fedora 24, Mac OS 10.11, Windows 10): 
**Platform** (e.g. 64-bit x86, 32-bit ARM): 
**OpenSlide version**: 
**Slide format** (e.g. SVS, NDPI, MRXS):

rmontroy:
This is a draft PR/RFC. I've not yet integrated the GDAL makefiles or configure scripts.

Fixes #200 

[Demo app](https://github.com/VanAndelInstitute/s3vs)


jaharkes:
I haven't looked at the details, but at a high level I would think openslide would not want to vendorize the gdal library as in this proposed patch. It looks like libgdal is already available on most Linux distributions and so it is easier to add as an external dependency like the many other libraries, glib, cairo, libxml, sqlite, libtiff, libjpeg, etc. that openslide already depends on.

It would also significantly reduce the size of the proposed pull request, from 164 files, to around 6 files, making it a lot easier to review.

jaharkes:
There could be a license issue with the libgdal dependency. Although the majority of GDAL is licensed as MIT/X style license (very permissive), some included files are licensed as BSD with an advertising clause which is as far as I know incompatible with OpenSlide's LGPL-2.1.

This issue is not made clear on https://gdal.org/license.html but is obvious once you read the [LICENSE.TXT](https://raw.githubusercontent.com/OSGeo/gdal/master/gdal/LICENSE.TXT) in their source.

rmontroy:
I'll make a custom GDAL fork for my needs and then redo this draft PR to have an optional dependency on GDAL.

bgilbert:
I discussed the license issue OOB with @jaharkes.  His concern was this clause from the `tif_float.c` license:

```
*       Redistributions in binary form must reproduce the above
copyright notice, this list of conditions and the following disclaimer
in the documentation and/or other materials provided with the
distribution.
```

That's actually not the advertising clause, it's the usual second clause from the BSD 3-clause license, which is GPL-compatible.  So that licensing issue shouldn't be a factor.

I agree that the library shouldn't be vendored here, though.  I'll comment further in #330.

PR #330, opened 2021-05-24
Support virtual file systems via GDAL

### LINKED ISSUE #200: Openslide can read the remote slide on internet?

rmontroy:
This is a draft PR/RFC.

Fixes #200 

[Demo app](https://github.com/VanAndelInstitute/s3vs)


bgilbert (commented):
Thanks for the PR!  My view from #200 hasn't changed, though; I'm still inclined to think we shouldn't do this.  There are a few reasons:

- Virtual filesystems are useful, but implementing one in each individual library or application seems like the wrong approach.  A better place to put a virtual filesystem is _in_ the filesystem, using FUSE.
- OpenSlide doesn't really optimize local I/O, so adding significant I/O latency could substantially affect performance.  We've had good results from a design where OpenSlide runs server-side instead of client-side, and serves tiles to remote clients over HTTP.  (Of course, that doesn't help if the tiles are stored in e.g. S3.)
- This change requires every filesystem access to be indirected, and it'd be fairly easy to miss cases without comprehensive tests.  For example, this PR doesn't address Sakura slides.
- I/O interception may not be possible for all slide formats in the future, depending on what utility libraries are needed to read the slide data.  For example, SQLite happens to provide a virtual filesystem API, but one could imagine a similarly complex library which did not.

Further notes on this implementation:
- OpenSlide is already closely tied to glib, and the associated gio library already includes a VFS, so if we were to support this functionality, gio seems like an obvious choice.  That would also avoid GDAL's not-really-URLs.
- This change alters the semantics of the path passed to `openslide_open()`, resulting in network access where none previously existed.  GDAL's path format might prevent accidental behavior changes, but not malicious ones; some applications that want to prevent unauthorized network access might now have to filter slide paths passed in from another component.  (I don't expect that to be a common case, but it might apply to someone.)  It seems better to provide a new open function for network-enabled access.

rmontroy:
Wow, that was a thorough review for a draft/RFC, thanks! I don't disagree with any of your points except one—"we shouldn't do this"—that you seemed to spell out yourself in your comments. The fact that OpenSlide I/O is not optimized combined with the network latency of a remote file makes use of S3 infeasible. S3 is a must nowadays, so the old argument against remote files is outdated, e.g., in the context of serverless applications.  
The network application (in this case, OpenSlide) really needs to be aware of network use by optimizing I/O. I think you're right that a new interface with different semantics should be made specifically for reading from remote files. This initial draft is admittedly a hack, but it was merely intended to have remote files reconsidered. I tried using Amazon EFS with AWS Lambda behind a VPC, and it's painful enough to make me attempt this route.  
BTW, the performance of this hack is great in terms of latency (no noticeable difference when viewing slides via OpenSeadragon) and the number of requests to S3. I was surprised that it worked as well as it did.

rmontroy:
Also, I can try to rework this into a new interface using gio/GVfs/GFile, if I can change your mind on it.

bgilbert:
Have you investigated S3-backed FUSE filesystems?  There are a number of existing implementations out there, and if none of them can do partial file requests, it shouldn't be too hard to build your own FUSE FS that does chunking and local caching.  The caching in particular might make the system more performant than anything we'd implement directly in OpenSlide.

PR #331, opened 2021-05-27
#38 Added functions to set cache size manually

### LINKED ISSUE #38: Allow cache customization
_Imported from Trac. This issue originally by *_agoode*\* at 2010-05-04T17:12:56Z.*

Right now, we have a simple per-slide cache that is not shared. There should be a way to allow a user to do caching themselves, or not at all if it is a one-pass run. Perhaps the user can provide a cache themselves, run uncached and cache everything themselves.

AlexanderVanEck:
I took the liberty of opening an MR for the changes @GeertLitjens was talking about in #38.

I'd appreciate a review/merge so I can also update openslide-python to be able to set this cache.
I will admit that I'm not very familiar in C and this is also not my code, please direct me on how to pass testing/linting checks.

@bgilbert would you be able to help me with this? :)

Thank you 🙇 

resolves #38

### DIFF
--- a/src/openslide.h
+++ b/src/openslide.h
@@ -99,6 +99,15 @@ openslide_t *openslide_open(const char *filename);
 OPENSLIDE_PUBLIC()
 int32_t openslide_get_level_count(openslide_t *osr);
 
+/**
+ * Set the cache size of the whole slide image.
+ *
+ * @param osr The OpenSlide object.
+ * @param level The cache size in bytes
+ * @since 3.3.0
+ */
+OPENSLIDE_PUBLIC()
+void openslide_set_cache_size(openslide_t *osr, int32_t cache_size);
 
 /**
  * Get the dimensions of level 0 (the largest level). Exactly
--- a/src/openslide.c
+++ b/src/openslide.c
@@ -408,6 +408,17 @@ int32_t openslide_get_level_count(openslide_t *osr) {
   return osr->level_count;
 }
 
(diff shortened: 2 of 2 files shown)


src/openslide.h
```diff
@@ -99,6 +99,15 @@ openslide_t *openslide_open(const char *filename);
 OPENSLIDE_PUBLIC()
 int32_t openslide_get_level_count(openslide_t *osr);
 
+/**
+ * Set the cache size of the whole slide image.
+ *
+ * @param osr The OpenSlide object.
+ * @param level The cache size in bytes
+ * @since 3.3.0
```
bgilbert, line 107: This would be 3.5.0.

bgilbert (commented):
The approach used here isn't thread-safe.  Resizing should happen inside the cache code, since it already has the proper locking.  Conveniently there are already functions for that purpose.  Also, if we're adding a function to set the cache size, there should probably be one to get the size as well.

However, as it says in https://github.com/openslide/openslide/issues/38#issuecomment-162705075, we should consider more general approaches to cache customization.  It'd be useful, for example, to allow sharing a single cache among multiple `openslide_t` handles.  That might be accomplished by allowing applications to provide their own cache implementation.

AlexanderVanEck:
Thank you @bgilbert for filling me in. I missed `_openslide_cache_set_capacity` and `_openslide_cache_get_capacity`. 🙌 

This PR can close and I'll follow-up with an issue for the suggested more general approaches to cache customization.

PR #333, opened 2021-06-11
Update openslide-vendor-mirax.c

GSz-3DH:
Dear OpenSlide Team!
We fixed this for You.
Bests,


src/openslide-vendor-mirax.c
```diff
@@ -1539,9 +1539,9 @@ static bool mirax_open(openslide_t *osr, const char *filename,
     goto FAIL;
   }
 
-  slidedat = _openslide_read_key_file(tmp, SLIDEDAT_MAX_SIZE,
-                                      G_KEY_FILE_NONE, err);
-  if (!slidedat) {
+  slidedat = g_key_file_new();
+  if (!_openslide_read_key_file(slidedat, tmp, SLIDEDAT_MAX_SIZE,
+                                G_KEY_FILE_NONE, err)) {
```
jaharkes, line 1544: It looks like you changed `_openslide_read_key_file` to take a `GKeyFile*` as the first argument, but this PR doesn't include the corresponding change to the function or other places (vendor-hamamatsu) where it is called from.

Without seeing the changes to `_openslide_read_key_file` I can't be really sure, but this change looks to be introducing a memory leak of the allocated GKeyFile object in the following error path. And if that cleanup happens in the changed function then there is serious mismatch of responsibilities which could easily lead to unintentional NULL pointer dereference errors when callers forget to check the `_openslide_read_key_file` error code.

src/openslide-vendor-mirax.c
```diff
@@ -511,7 +511,7 @@ static bool read_nonhier_record(FILE *f,
   }
 
   // read pagesize == 1
-  if (read_le_int32_from_file(f) != 1) {
+  if (read_le_int32_from_file(f) < 1) {
     g_set_error(err, OPENSLIDE_ERROR, OPENSLIDE_ERROR_FAILED,
                 "Expected 1 value");
```
jaharkes, line 516: The preceding comment and error output no longer match the actual test.

When is this value larger than 1? Is there range of possible values or can it be as much as MAX_UINT? How do we know other code doesn't need changes to handle these larger values.

jaharkes: It looks like PR #293 has the same `!=1` to `< 1` change and documents that it seems to occasionally be 2.
That PR also changes some aperio slide parsing, which seems unrelated to the mirax change.

src/openslide-vendor-mirax.c
bgilbert: Also, even if the above problems are fixed, it'd be helpful to know the purpose of the change.

GSz-3DH: It was a mistake, the source was downloaded from the website, not github.

GSz-3DH (commented):
This works for us.

GSz-3DH:
Hi,
It would be nice if You could recheck the mod:
https://github.com/3dhistech/openslide/commit/eb76d8fe3e9225a53d8511b9be313cb74011f621
That's all, the other file was not modified by us, it was from an early source of openslide.
Bests,
Gergely

jjhbw:
Our lab is having trouble opening some MRXS slides, which is supposedly fixed by this PR. See issue #251.
We can test this fix on our data to see if it works? If it does work, what else would be needed for this to get merged?

GSz-3DH:
Dear All,

MRXS to MRXS conversion is not just a bandaid, but the solution for You.

There are a few more things happening during the mrxs to mrxs conversion:
auto color adjustment for FL slides,
bilinear filter for smoothing,
pixel filter for smoothing,
intensity stitching to make overall intensity balanced,
stitching filter to smooth stitched edges,
optimize to slow connection is basically recompressing the image in jpg80, which can also smooth the image.
And maybe that's not all, I'm not sure.
We have these filters built-in to our viewers, that is why an overlapped mrxs image has problems in all alternative viewers.

I guess openslide will not do the above image manipulation ever.

So, the solution is to do mrxs to mrxs conversion, and use de-everlapped images.
The conversion may be applied during scanning.
local scanning DX software>Turn on: Change file format (MRXS)
server scanning DX software>Turn on: Apply slide processing (MRXS)
The RX and the P1000 has similar options.

And You can always use the folder watch feature of SlideMaster, and convert the scanned slides automatically.
This works with any 3dh scanner version.

Kind regards,

bgilbert:
Closing in favor of https://github.com/openslide/openslide/pull/521.  Thanks for the PR.

PR #335, opened 2021-07-04
Add exported cache management API

### LINKED ISSUE #38: Allow cache customization

bgilbert:
Create a slide handle with a default cache, but allow the user to separately allocate a cache of any size (including zero) and attach it to one or more slide handles.  This allows sharing a cache between slides.  The public API is:

```C
// allocate new cache
openslide_cache_t *openslide_cache_create(int64_t capacity);

// set cache for slide handle, replacing the existing cache
void openslide_set_cache(openslide_t *osr, openslide_cache_t *cache);

// release application's reference to a cache; cache is not freed until
// all attached slides are detached or closed
void openslide_cache_release(openslide_cache_t *cache);
```

For now, don't support resizing an existing cache, nor external caches implemented by the application.  Those features can be added within the existing API if they turn out to be needed.

Fixes #38.

Still needed:
- [x] Tests

### DIFF
--- a/src/openslide-cache.c
+++ b/src/openslide-cache.c
@@ -48,7 +48,7 @@ struct _openslide_cache_value {
 struct _openslide_cache_entry {
   gint refcount;  // atomic ops only
   void *data;
-  int size;
+  uint64_t size;
 };
 
 struct _openslide_cache {
@@ -57,8 +57,8 @@ struct _openslide_cache {
   GHashTable *hashtable;
   int refcount;
 
-  int capacity;
-  int total_size;
+  uint64_t capacity;
+  uint64_t total_size;
 
   gint warned_overlarge_entry;
 };
@@ -72,11 +72,11 @@ struct _openslide_cache_binding {
 
 // eviction
 // mutex must be held
-static void possibly_evict(struct _openslide_cache *cache, int incoming_size) {
-  g_assert(incoming_size >= 0);
-
-  int size = cache->total_size + incoming_size;
-  int target = cache->capacity;
+static void possibly_evict(struct _openslide_cache *cache,
+                           uint64_t incoming_size) {
+  uint64_t size = cache->total_size + incoming_size;
+  uint64_t target = cache->capacity;
+  g_assert(size > cache->total_size);
 
   while(size > target) {
     // get key of last element
@@ -125,8 +125,8 @@ static void hash_destroy_value(gpointer data) {
   g_queue_delete_link(value->cache->list, value->link);
 
   // decrement the total size
+  g_assert(value->entry->size <= value->cache->total_size);
   value->cache->total_size -= value->entry->size;
-  g_assert(value->cache->total_size >= 0);
 
   // unref the entry
   _openslide_cache_entry_unref(value->entry);
@@ -135,7 +135,7 @@ static void hash_destroy_value(gpointer data) {
   g_slice_free(struct _openslide_cache_value, value);
 }
 
-struct _openslide_cache *_openslide_cache_create(int capacity_in_bytes) {
+struct _openslide_cache *_openslide_cache_create(uint64_t capacity_in_bytes) {
   struct _openslide_cache *cache = g_slice_new0(struct _openslide_cache);
 
   // init mutex
@@ -224,7 +224,7 @@ void _openslide_cache_put(struct _openslide_cache_binding *cb,
 			  int64_t x,
 			  int64_t y,
 			  void *data,
-			  int size_in_bytes,
+			  uint64_t size_in_bytes,
 			  struct _openslide_cache_entry **_entry) {
   // always create cache entry for caller's reference
   struct _openslide_cache_entry *entry =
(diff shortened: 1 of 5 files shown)


jaharkes:
I think the int64_t change could be split out, it is really an independent change. Also, I don't see how those values could go negative unless there are code errors (caching data from a failed operation that returned an error ssize_t?). So I think using size_t in function parameters is still a good idea either way along with some better testing/asserts to catch integer overflows during eviction handling.

If there always is a 1:1 between the openslide handle and the cache binding, I don't think it is necessary to allocate a separate structure for the binding.

Caches can be shared between different slides, so the cache key probably needs a differentiator to indicate not just the plane and coordinates, but also an (opaque) reference to the slide object.

Also not sure about the need to drop values from a binding when detaching. It is an optimization because those values should LRU expire automatically as new values are added to the cache. If it is because each cache key nees a reference to the openslide handle, then it may be possible to just iterate over all cached values and drop based matching the reference and avoid the overhead of extra pointers and lists in the cache value and binding structure for this infrequent cleanup case.

bgilbert:
Thanks for the review!

> I think the int64_t change could be split out, it is really an independent change.

Well, sure.

> Also, I don't see how those values could go negative unless there are code errors (caching data from a failed operation that returned an error ssize_t?). So I think using size_t in function parameters is still a good idea either way along with some better testing/asserts to catch integer overflows during eviction handling.

I originally had it as `size_t`, but in general the codebase avoids unsigned integers and architecture-dependent sizes.  The size calculations performed by the vendor drivers all operate on signed integers for that reason.  One option is to use `size_t` in the external API but `int64_t` within the caching code.

> ~If there always is a 1:1 between the openslide handle and the cache binding, I don't think it is necessary to allocate a separate structure for the binding.~ Actually thinking about this more, this is quite similar to the Arc in Rust, and it makes it easier to replace this with an alternative implementation.

The idea was to keep the cache code self-contained.  If we don't have a separate structure for the binding, then the cache getter/putter needs to take a slide handle parameter, and acquire a lock embedded directly in the `openslide_t`.

> Caches can be shared between different slides, so the cache key ~probably~ needs a differentiator to indicate not just the plane and coordinates, but also an (opaque) reference to the slide object.

The current code is implicit, but safe.  The plane is always a pointer to a struct owned by the slide handle.  We could add a `openslide_t` pointer to the cache key, but it'd really only be to guard against the case that some vendor driver incorrectly passes a pointer to a plane struct it doesn't own.

> Also not sure about the need to drop values from a binding when detaching. It is an optimization because those values should LRU expire automatically as new values are added to the cache. If it is because each cache key needs a reference to the openslide handle, then it you can just iterate over all cached keys and drop based on the openslide handle reference and avoid the overhead of storing and maintaining extra pointers and lists in the cache value and binding structures for this infrequent cleanup case.

Yeah, it's completely an optimization, and I was on the fence about including it.  The time and space overhead of one list entry doesn't seem that significant, though?

bgilbert:
This PR doesn't try to share cache values across multiple slide handles.  If there are multiple slide handles referencing the same slide file, or different files with the same quickhash, a shared cache will keep one copy of a tile per slide handle.  Changing that would require:

- Tracking the file path or quickhash in the cache value.  The quickhash should be unique but may not always exist; we could fall back to the file path (or inode?) in that case.
- Replacing the opaque plane pointer with a data structure that can be compared for equality.  This could just be a stringified description of the plane.

For both parts we might want to have a refcounted string interning pool, to avoid doing a lot of string comparisons.

I didn't pursue this approach here because we [document](https://openslide.org/api/openslide_8h.html#a6d49bdccc0883d8cfc7d9e8085384b6d) that `openslide_open()` is expensive and that applications should maintain a cache of slide handles, so the gain seemed minimal.

bgilbert:
Actually, as written, the cache invalidation is needed for correctness.  The slab allocator might recycle an old plane pointer in a future slide handle, at which point, if we didn't have invalidation, we could return stale cache entries.

Alternatively, we could drop the cache invalidation code and give each cache binding a generation number which is included in the cache key.

jaharkes:
I think you have addressed all my comments, but I'll just resummarize my thinking.

I would prefer using unsigned integers in at least the public API. In fact assuming we'll never cache chunks larger than 4GB it could be a uint32 and there would be no worry about accidentally overflowing the internal 64-bit counters.

The separate cache binding is useful to keep the cache related code separate and clearly identifies what is being protected by the mutex.

As far as the pruning of the cache when it is detached is concerned, ideally the unnecessary cached data would just age out. As you noted however, although the pointer to the plane is an opaque value, there is a reasonable chance it could be reused between different slides which could lead to pulling stale data from the cache.

I personally don't like adding extra linked lists and mutexes, which adds a little bit of overhead on every operation to cover the infrequent detach from cache case. One alternative I can think of is to iterate through all keys and match the plane pointers which could be more costly. Or as you suggested add a generation number to the binding and cache keys to avoid returning stale data which I believe is an API change for lookups, but it avoids runtime overhead and lets old entries expire naturally.

bgilbert:
Updated to use `size_t` in the public API and `uint64_t` in the cache code.  `uint64_t` turned out to be pretty clean, and it keeps architecture-dependent sizes out of the core.

I also dropped the cache invalidation code in favor of including a binding identifier in cache keys.

bgilbert:
Added tests; dropped RFC.

jaharkes (approved):
LGTM

Issue #297, opened 2020-07-10
No longer able to retrieve label from SVS Format (Leica, formally Aperio)


MarkusRissmann:
## Context

**Issue type** Feature request (update of a feature that no longer works)
**Operating system** Windows
**Platform** x64
**OpenSlide version**: latest
**Slide format** newer SVS

## Details

I no longer can retrieve the label using the latest openslide library. I am pretty sure that Leica (formally Aperio has probably changed the the format a little.) Everything else works though.

Ok to post here or should I use the mailing list?

phillipjohnson:
HI @MarkusRissmann I am curious what scanner model you're using? We also have this problem with some of their newer equipment. I got this info from Leica:

> On the GT450, the Image Description value for macro and label layers empty. To detect the macro and
> label layers, it is recommended to use the TIFFTAG_SUBFILETYPE tag. The macro layer will always have a
> value of 9, and the label layer will always have a value of 1. This applies to both images generated by the
> AT2 and the GT450

I don't have a workaround yet as I believe a change to [openslide-vendor-aperio.c#add_associated_image](https://github.com/openslide/openslide/blob/main/src/openslide-vendor-aperio.c#L338) is required.

fiendish:
It seems like https://openslide.org/formats/aperio/ should be updated to remove or amend the parts that say `the name “label” is given in ImageDescription` and `the name “macro” is given in ImageDescription`.

nirbenor-ibex:
> HI @MarkusRissmann I am curious what scanner model you're using? We also have this problem with some of their newer equipment. I got this info from Leica:
> 
> > On the GT450, the Image Description value for macro and label layers empty. To detect the macro and
> > label layers, it is recommended to use the TIFFTAG_SUBFILETYPE tag. The macro layer will always have a
> > value of 9, and the label layer will always have a value of 1. This applies to both images generated by the
> > AT2 and the GT450
> 
> I don't have a workaround yet as I believe a change to [openslide-vendor-aperio.c#add_associated_image](https://github.com/openslide/openslide/blob/main/src/openslide-vendor-aperio.c#L338) is required.

Thanks! See: https://github.com/openslide/openslide/pull/343

MarkusRissmann:
Correct, it is the GT450.

PR #343, opened 2021-10-17
Support Leica GT450 scanner associated images

nirbenor-ibex:
Solves: https://github.com/openslide/openslide/issues/297

### DIFF
--- a/src/openslide-vendor-aperio.c
+++ b/src/openslide-vendor-aperio.c
@@ -536,7 +536,27 @@ static bool aperio_open(openslide_t *osr,
       }
     } else {
       // associated image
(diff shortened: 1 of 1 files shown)


bgilbert:
> So far I've been suspecting that the format spec did not actually change and that the original detection mechanism here was someone just winging it

Aperio's 2008 spec document (_Digital Slides and Third-Party Data Interchange_) described the label and macro images but did not suggest a detection mechanism.  However you appear correct that the subfile type has always been present.  In #730 I've switched entirely to detecting via the subfile type, which also allows the code to be cleaned up a bit.

Closing in favor of #730.  Thanks for the PR!

PR #349, opened 2022-04-25
Set MPP properties for generic TIFF

Falcury:
For generic TIFF files, the `openslide.mpp-x` and `openslide.mpp-y` properties are not being set, even if the TIFF file contains XResolution and YResolution tags. 

Because of this, you could miss the resolution information for generic TIFF files if you only checked for `openslide.mpp-x` and `openslide.mpp-y` and not for `tiff.XResolution` and `tiff.YResolution`. 

The changes in this PR address this limitation. The added code for setting the `openslide.mpp-x` and `openslide.mpp-y` properties is copied from the way these same properties are set in `openslide-vendor-leica.c`:
https://github.com/openslide/openslide/blob/2d5b9aa5d87ad8a6f9e6c47e5423c5e421a39e9c/src/openslide-vendor-leica.c#L332-L344
https://github.com/openslide/openslide/blob/2d5b9aa5d87ad8a6f9e6c47e5423c5e421a39e9c/src/openslide-vendor-leica.c#L853-L860

### DIFF
--- a/src/openslide-vendor-generic-tiff.c
+++ b/src/openslide-vendor-generic-tiff.c
@@ -188,6 +188,20 @@ static int width_compare(gconstpointer a, gconstpointer b) {
   }
 }
 
+static void set_resolution_prop(openslide_t *osr, TIFF *tiff,
+                                const char *property_name,
+                                ttag_t tag) {
+  float f;
+  uint16_t unit;
+
+  if (TIFFGetFieldDefaulted(tiff, TIFFTAG_RESOLUTIONUNIT, &unit) &&
+      TIFFGetField(tiff, tag, &f) &&
+      unit == RESUNIT_CENTIMETER) {
+    g_hash_table_insert(osr->properties, g_strdup(property_name),
+                        _openslide_format_double(10000.0 / f));
+  }
+}
+
 static bool generic_tiff_open(openslide_t *osr,
                               const char *filename,
                               struct _openslide_tifflike *tl,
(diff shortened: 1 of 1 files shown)


kaczmarj:
+1 this would be nice to have in openslide. our group often works with generic, multi-resolution TIFFs (converted from proprietary formats), and i have always noticed that mpp is missing despite the x and y resolution being stored in the TIFF file.

Falcury:
Rebased and added the DCI sign-off message in the commit.

bgilbert:
Thanks for the PR!  I think we didn't do this initially because those TIFF tags might be nonsense.  The [property docs](https://openslide.org/docs/properties/) already say "may not be present or accurate", though, so it probably makes sense to add this.

src/openslide-vendor-generic-tiff.c
```diff
@@ -175,6 +175,20 @@ static int width_compare(gconstpointer a, gconstpointer b) {
...
 
+static void set_resolution_prop(openslide_t *osr, TIFF *tiff,
+                                const char *property_name,
+                                ttag_t tag) {
+  float f;
+  uint16_t unit;
+
+  if (TIFFGetFieldDefaulted(tiff, TIFFTAG_RESOLUTIONUNIT, &unit) &&
+      TIFFGetField(tiff, tag, &f) &&
+      unit == RESUNIT_CENTIMETER) {
```
bgilbert, line 186: Since this is a generic driver, we should probably also handle the conversion if the unit is inches.

src/openslide-vendor-generic-tiff.c
```diff
@@ -175,6 +175,20 @@ static int width_compare(gconstpointer a, gconstpointer b) {
   }
 }
 
+static void set_resolution_prop(openslide_t *osr, TIFF *tiff,
```
bgilbert, line 178: Rather than duplicating the code from Leica, let's move this into the generic TIFF decoder and have both drivers call it.  In cases where either `tiff` or `tifflike` could provide some functionality, we usually put it into `tifflike`, so let's make it a helper function there.  But for now I'd say let's _not_ make it happen automatically from `store_and_hash_properties`, since there are lots of formats that record the MPP in unusual ways, and it's better for drivers to configure the MPP explicitly.

src/openslide-vendor-generic-tiff.c
```diff
@@ -247,13 +261,23 @@ static bool generic_tiff_open(openslide_t *osr,
 
   // set hash and properties
   struct level *top_level = level_array->pdata[level_array->len - 1];
+  int32_t property_dir = 0;
```
bgilbert, line 264: Unnecessary variable?

bgilbert:
I've rebased this onto `main`, but am closing it in favor of #516, which implements the approach requested above.

Issue #355, opened 2022-06-14
Require contributors to assent to the Developer Certificate of Origin


bgilbert:
As part of the ongoing effort to modernize OpenSlide project infrastructure and make it easier to contribute, we should formalize our protections against contributions that may be legally encumbered, including contributions based on file format documentation obtained under an NDA or a restrictive license.

Some projects accomplish this via a Contributor License Agreement, but that approach is controversial.  CLAs typically require contributors to either assign their copyright to the project, or at least grant the project an unlimited right to relicense the code, allowing it to be used in proprietary products.  I'm among the people who believe that open-source projects are healthier when their copyrights are held by a variety of people, ensuring that relicensing the project is difficult or impossible.

Instead, we can use the [Developer Certificate of Origin](https://developercertificate.org/), originated by the Linux kernel and also used by other projects.  It would protect the project by requiring contributors to certify that they have the right to submit their contribution, but it does not grant the project any additional rights, nor require the contributor to give up any.  Making this certification would be a lightweight process: contributors would just add `Signed-off-by: Developer Name <developer.email@example.com>` to their commit messages.

DCO paragraph (a) prohibits contributions that the contributor does not have the right to submit, which would provide the desired protection against code derived from encumbered documentation.  Since this point is crucial for OpenSlide, I think the contribution guide and CI notifications should explicitly call it out to avoid any confusion.

We can implement the new requirement by:

- [x] Adding a contribution guide with the full text of the DCO and instructions for signing off on commits.
- [x] Adding CI checks that fail if commits are not properly signed off.
- [x] [Requiring signoff on web-based commits](https://github.blog/changelog/2022-06-08-admins-can-require-sign-off-on-web-based-commits/).
- [x] Strengthening GitHub branch protection so that everyone, including repository maintainers, must submit a PR that passes CI before landing code.  (In practice, this would affect mostly me.)

To be clear, this change isn't the result of any particular incident.  Potentially-encumbered code has been a concern since the beginning of the project, and it's been clear for a while that we should formalize our existing processes for keeping such material out of OpenSlide.

I'll open a PR soon in the [.github metadata repo](https://github.com/openslide/.github) with the exact CONTRIBUTING text and CI checks I'm proposing, and will link the PR here for visibility.

bgilbert:
Documentation and CI jobs are posted for review:
- https://github.com/openslide/openslide.github.io/pull/18
- https://github.com/openslide/.github/pull/1

bgilbert:
[...]

PR #359, opened 2022-06-23
workflows: add DCO verification

bgilbert:
See https://github.com/openslide/openslide/issues/355.

Requires https://github.com/openslide/.github/pull/1.

### DIFF
--- a/.github/workflows/dco-check.yml
+++ b/.github/workflows/dco-check.yml
@@ -1,0 +1,13 @@
+name: DCO
+
+on:
+  pull_request:
+    branches: [main]
+
+permissions:
+  contents: read
+
+jobs:
+  check:
+    name: Organization
+    uses: openslide/.github/.github/workflows/dco-check.yml@main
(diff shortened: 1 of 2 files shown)


.github/workflows/dco-check.yml
```diff
@@ -0,0 +1,13 @@
...
+  pull_request:
+    branches: [main]
+
+permissions:
+  contents: read
+
+jobs:
+  check:
+    name: Organization
+    uses: openslide/.github/.github/workflows/dco-check.yml@main
```
jaharkes, line 13: any reason this is not just openslide/.github/workflows/dco-check.yml?

Oh wait that would be this file. I really have to figure out github actions better.

bgilbert: I thought it'd be cleaner to have the individual repos all invoke workflows from a separate org-level metadata repo, so we don't have redundant copies everywhere and don't have the main OpenSlide repo doing double duty as a library of CI workflows.

Issue #309, opened 2020-09-15
Consider unpacking test xdeltas onto reflinked base file


bgilbert:
To dramatically reduce the disk space requirements for a full test run, consider reflinking the base file first, then reading `xdelta3`'s stdout and conditionally overwriting blocks of the reflinked file.  This will only help on XFS and btrfs.

PR #362, opened 2022-07-05
test/driver: unpack xdelta onto reflinked base file, if possible

bgilbert:
If the underlying filesystem supports it (XFS or btrfs), use reflinks to reuse unmodified data blocks from the pristine copy of the file.  We could do this at blksize granularity (4 KiB), but save some syscalls by doing it in 1 MiB chunks.  In practice, most of our xdeltas only apply a handful of distinct modifications to a file; we're mainly trying to avoid copying hundreds of MiB of bulk data.

This reduces `_slidedata` from 124 GiB to 22 GiB on XFS.

Also avoid fetching zipped slides into `/tmp`, since that's often a `tmpfs`.

Closes https://github.com/openslide/openslide/issues/309.


bgilbert:
(APFS benefits from this too.)

PR #378, opened 2022-08-06
synthetic: test decoding corrupt JPEG and PNG

bgilbert:
Both decoders use `setjmp`/`longjmp` for error handling, which has historically had problems on 64-bit Windows with MinGW.  Let's include it in our smoke tests.

For https://github.com/openslide/openslide-winbuild/pull/49.

### DIFF
--- a/misc/make-synthetic-item.py
+++ b/misc/make-synthetic-item.py
@@ -10,6 +10,7 @@ for name in sys.argv[1:]:
     print(f'  &(const struct synthetic_item){{')
     print(f'    .name = "{name}",')
     print(f'    .description = "fill this in",')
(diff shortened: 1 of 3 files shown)


bgilbert:
The 64-bit test isn't failing as expected.  At a guess, we'd need the library to be invoked from an MSVC-compiled program to reproduce the issue.  Closing for now.

bgilbert:
It appears we have a `setjmp` workaround for 64-bit Windows (https://github.com/openslide/openslide-winbuild/pull/49) so this check is still useful to make sure the workaround doesn't regress the pure-MinGW case.  Reopening.

PR #382, opened 2022-08-12
JPEG and PNG cleanups; fix JPEG error path when compiled with clang

bgilbert:
Fixes macOS CI.  See individual commits for details.


bgilbert:
@jaharkes pointed out that the clang behavior actually seems to be a bug.  Reported in https://github.com/llvm/llvm-project/issues/57110.

PR #286, opened 2020-01-08
Add Support for Huron Digital Pathology Scanner Generated Images

w8choi:
This version adds supports for the huron scanner generated images.

Issue #260, opened 2019-01-17
Add Support for Huron Digital Pathology (HDP) images


seshah:
## Context

**Issue type** (bug report or feature request): feature request
**Operating system** (e.g. Fedora 24, Mac OS 10.11, Windows 10): Windows 7
**Platform** (e.g. 64-bit x86, 32-bit ARM):  64-bit x86
**OpenSlide version**: 3.4.1
**Slide format** (e.g. SVS, NDPI, MRXS): SVS, TIF 

## Details

Hello,

I recently started working with HDP. They're a company that scans images and saves them as .tif files (which they sometimes convert to .svs).  The .tif file format is very similar to that of the Aperio .tif file format.  However, since the vendor isn't Aperio, when I tried opening them using openslide, they were being considered as generic-tiff images with a single level, even though they were tiled pyramids of an image.  I kind of hacked the openslide-vendor-aperio.c file to make it work with HDP images.  

The main changes I made are  : 
 - added around line 45
```
static const char APERIO_DESCRIPTION[] = "Aperio"; 
///////////////////////////////////////////////////////////////////// my edit begins
static const char HURON_DESCRIPTION[] = "Huron";
///////////////////////////////////////////////////////////////////// my edit ends
```
- added around line 302
```
if (!g_str_has_prefix(tagval, APERIO_DESCRIPTION)) {
//////////////////////////////////////////////////////////////////////////// my edit begins
    const char *tagval2 = _openslide_tifflike_get_buffer(tl, 0, TIFFTAG_MAKE, err);
   if (tagval2) {
    	if (g_str_has_prefix(tagval2, HURON_DESCRIPTION) ){
    		return true;
    	}
    }
//////////////////////////////////////////////////////////////////////////// my edit ends
     g_set_error(err, OPENSLIDE_ERROR, OPENSLIDE_ERROR_FAILED,
                    "Not an Aperio slide");
      return false;
  }
```
other changes were g_message statements that I used for debugging/testing purposes. 

**Would it be possible for you to add support for HDP images?**

I'm in the process of getting permission to send you a HDP image if needed, for now though I've attached 2 simple log files that have the output of the properties of three images : 
-first image is an SVS image that isn't from HDP
-second image is an HDP  TIF image
-third image is an HDP  SVS image
HDP scanners can make Uncompressed, JPG compressed and JPG2000 compressed images.  

[...]

PR #398, opened 2022-10-14
Add Support for Huron Digital Pathology Scanner Generated Images

101x4:
This version adds supports for the huron scanner generated images.
Please close the former PR (https://github.com/openslide/openslide/pull/286), which is created by former employee from Huron, thanks.


bgilbert:
@101x4 It looks like there are two problems with your signoff lines: the signoffs need to exactly match the name and email address of the commit author, and the first two commits in the PR have not been signed off.  You'll need to edit your existing commit messages with `git rebase -i` and then force-push.

bgilbert:
For #260.

bgilbert:
@101x4 Use `git rebase -i origin/main` to squash all of the commits down to one commit with the correct signoff line, then `git push -f`.

101x4:
I didn't meet the error of "src/openslide-vendor-huron.c:450:18: error: incompatible types when initializing type ‘TIFF *’ {aka ‘struct tiff *’} using type ‘struct _openslide_cached_tiff’" when I was running 'autoreconf -i' myself. Any advice? Thanks.

uoffv:
I think the problem is that Openslide has changed a lot and our modification for HDP several years ago is no longer effective :(

101x4:
@bgilbert Can you help with this error: "Fetching zlib..." failed with "Failed: ${wget} -P tar "$url" (line 255)"?
I checked the Zlib website and it showed that we need to use version of 1.2.13. Thanks.
image

bgilbert:
It seems the previous release of zlib has been removed from its website for some reason.  https://github.com/openslide/openslide-winbuild/pull/54 should fix CI.

101x4:
@bgilbert Fetching openslide-java also failed with "Failed: ${wget} -P tar "$url" (line 255)". And how can I rerun the checks?

bgilbert:
Thanks for the PR!  Could you [submit some sample files](https://forms.gle/Emr38bepoqC4wUqk6) that we can include in our sample library and in automated test cases?  There's a question at the bottom of the form asking if we can redistribute the samples; we'd need that permission in order to use the slides for these purposes.

101x4:
Images uploaded, and please ignore the zip file. Thanks.

bgilbert:
Thanks for the PR and especially for the sample slides.  This PR includes some indirections and workarounds based on the Aperio driver which shouldn't be necessary for Huron, so I've opted for a fresh implementation in https://github.com/openslide/openslide/pull/734 which will be included in the next release.  I appreciate your patience through this very long process.

#260 reports that Huron slides support JPEG 2000 compression, but I don't have any samples that use it.  Rather than retain code from the Aperio driver that may not apply here (e.g. does Huron support both 33003 and 33005?) I've opted to omit JPEG 2000 support for now.  If we [receive](https://openslide.org/submit/) a redistributable Huron sample with JPEG 2000 compression I'll be happy to re-add support in the new implementation.

PR #405, opened 2022-11-04
adding support for optra otif images

surajoptra:
added driver for otif and modified supporting files.

Signed-off-by: surajoptra <s.somwanshi@optraventures.com>


bgilbert:
Thanks for the PR!  Could you [submit some sample files](https://forms.gle/Emr38bepoqC4wUqk6) that we can include in our sample library and in automated test cases?  There's a question at the bottom of the form asking if we can redistribute the samples; we'd need that permission in order to use the slides for these purposes.

surajoptra:
Submitted sample file. Please verify.

Best Regards,
Suraj S.

PR #407, opened 2022-11-18
add JPEG XR library to github CI

iewchen:
Add JPEG XR to CI so that it can build zeiss driver.
    
- jxrlib from homebrew has a bug in JXRMeta.h. Needs replace with sed.
    
Only tested Linux and MacOS. It may not work on Windows.

Signed-off-by: Wei Chen <chenw1@uthscsa.edu>

### DIFF
--- a/.github/workflows/c.yaml
+++ b/.github/workflows/c.yaml
@@ -56,6 +56,7 @@ jobs:
                 jpeg-turbo \
                 libtiff \
                 openjpeg \
+                jxrlib \
                 gdk-pixbuf \
                 libxml2 \
                 sqlite \
@@ -63,6 +64,8 @@ jobs:
                 glib \
                 xdelta \
                 doxygen
+            # jxrlib from homebrew has a bug
+            LC_ALL=C sed -i '' 's/#endif UNREFERENCED_PARAMETER/#endif/' /usr/local/Cellar/jxrlib/1.1_1/include/jxrlib/JXRMeta.h
             python -m pip install --upgrade pip
             pip install requests PyYAML
             ;;
@@ -78,6 +81,7 @@ jobs:
                     dnf copr enable -y bgilbert/el8-pixman-openslide
                     pyver=38
                     python=python38
+                    dnf install -y epel-release
                     ;;
                 9)
                     dnf install -y 'dnf-command(config-manager)'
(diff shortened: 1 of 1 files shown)


.github/workflows/c.yaml
```diff
@@ -59,13 +61,16 @@ jobs:
...
                 openjpeg \
+                jxrlib \
                 gdk-pixbuf \
                 libxml2 \
                 sqlite \
                 cairo \
                 glib \
                 xdelta \
                 doxygen
+            # jxrlib from homebrew has a bug
```
bgilbert, line 72: Could you file a Homebrew issue or PR for this and add a link to it in a comment here?

.github/workflows/c.yaml
```diff
@@ -81,16 +86,22 @@ jobs:
...
                     pyver=38
                     python=python38
+                    dnf install -y epel-release
+                    dnf install -y jxrlib-devel
                     ;;
                 9)
                     dnf install -y 'dnf-command(config-manager)'
                     dnf config-manager --set-enabled crb
                     dnf install -y epel-release epel-next-release
+                    # jxrlib is not in EPEL 9 as of Nov 17, 2022
```
bgilbert, line 96: It is now.

.github/workflows/c.yaml
```diff
@@ -59,13 +61,16 @@ jobs:
...
+                jxrlib \
                 gdk-pixbuf \
                 libxml2 \
                 sqlite \
                 cairo \
                 glib \
                 xdelta \
                 doxygen
+            # jxrlib from homebrew has a bug
+            LC_ALL=C sed -i '' 's/#endif UNREFERENCED_PARAMETER/#endif/' /usr/local/Cellar/jxrlib/1.1_1/include/jxrlib/JXRMeta.h
```
bgilbert, line 73: Hmm, why is `LC_ALL=C` required here?

bgilbert (commented):
This looks good; just a couple small fixes.  Windows builds are handled separately; `build.sh` in [this repo](https://github.com/openslide/openslide-winbuild) would need to be updated to build jxrlib.

.github/workflows/c.yaml
iewchen: There is a "illegal byte sequence" error in that sed command. LC_ALL=C fixes it.

.github/workflows/c.yaml
iewchen: Please see https://github.com/Homebrew/homebrew-core/issues/117415 .

iewchen:
Thanks for the review! I am closing this PR because its changes is part of #396.

PR #413, opened 2022-12-08
Drop deprecated functions, implement other pending ABI breaks

bgilbert:
Remove deprecated functions and implement cleanups that require breaking changes.

Removed functions:

- `openslide_cancel_prefetch_hint` - deprecated since 2010; has always done nothing
- `openslide_can_open` - deprecated since 2014; use `openslide_detect_version` instead
- `openslide_get_best_layer_for_downsample` - deprecated since 2012; use `openslide_get_best_level_for_downsample` instead
- `openslide_get_comment` - deprecated since 2012; use `openslide_get_property_value(osr, OPENSLIDE_PROPERTY_NAME_COMMENT)` instead
- `openslide_get_layer0_dimensions` - deprecated since 2012; use `openslide_get_level0_dimensions` instead
- `openslide_get_layer_count` - deprecated since 2012; use `openslide_get_level_count` instead
- `openslide_get_layer_dimensions` - deprecated since 2012; use `openslide_get_level_dimensions` instead
- `openslide_get_layer_downsample` - deprecated since 2012; use `openslide_get_level_downsample` instead
- `openslide_give_prefetch_hint` - deprecated since 2010; has always done nothing

Other breaking changes:

- Change the capitalization of the `G_LOG_DOMAIN` from `Openslide` to `OpenSlide`.  This is an ABI change for callers that register glib log handlers to control debug logging, though since OpenSlide 3.3.0, we only use glib logging for programming errors and debug messages.  Applications that want to be compatible with both old and new versions should register handlers for both `Openslide` and `OpenSlide`.
- Change `openslide_read_associated_image()` to clear the `dest` buffer on error, rather than leaving it unmodified.  The old behavior was documented, but most applications probably don't rely on it.

This will require a major version bump (4.0.0) and a soname bump.  Those aren't implemented by this PR, but will be done at release time.  The soname bump will require changes to any library user that hardcodes the OpenSlide library filename, including OpenSlide Python and openslide-winbuild.

### DIFF
--- a/src/openslide.h
+++ b/src/openslide.h
@@ -477,7 +477,7 @@ void openslide_read_associated_image(openslide_t *osr,
  *
  * @param capacity The capacity of the cache, in bytes.
  * @return A new cache.
- * @since 3.5.0
+ * @since 4.0.0
  */
 OPENSLIDE_PUBLIC()
 openslide_cache_t *openslide_cache_create(size_t capacity);
@@ -488,7 +488,7 @@ openslide_cache_t *openslide_cache_create(size_t capacity);
  *
  * @param osr The OpenSlide object.
  * @param cache The cache to attach.
- * @since 3.5.0
+ * @since 4.0.0
  */
 OPENSLIDE_PUBLIC()
 void openslide_set_cache(openslide_t *osr, openslide_cache_t *cache);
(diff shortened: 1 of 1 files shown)


bgilbert:
Additional ABI break (for some undocumented functionality) in https://github.com/openslide/openslide/pull/472.

PR #425, opened 2023-01-14
Convert build system from Autotools to Meson

### LINKED ISSUE #316: Consider switching the build to Meson
Autotools has historically been the default for Linux systems libraries, but it's admittedly pretty awkward, and recently a bunch of projects (including some of our dependencies) have been moving to Meson.  Consider switching.

bgilbert:
Autotools has historically been the default for Linux system libraries, but it's pretty awkward by modern standards.  Switch to Meson, which is more performant, simpler to develop for, and increasingly common in the Linux space.

Thanks to @jaharkes for doing the initial conversion work!

Fixes #316.


meson.build
```diff
@@ -0,0 +1,250 @@
...
+xml_dep        = dependency('libxml-2.0')
+sqlite_dep     = dependency('sqlite3', version : '>=3.14')
+valgrind_dep   = dependency('valgrind', required : false)
+
+doc = get_option('doc')
+if not doc.disabled()
+  doxygen = find_program('doxygen', required : doc)
+else
+  doxygen = disabler()
+endif
```
eli-schwartz, line 141: Hmm, can't you just use:
```suggestion
doxygen = find_program('doxygen', required : get_option('doc'), disabler: true)
```

scripts/dist.py
```diff
@@ -0,0 +1,16 @@
...
+from pathlib import Path
+import shutil
+import subprocess
+
+base = Path(os.getenv('MESON_DIST_ROOT'))
+
+base.joinpath('.gitattributes').unlink()
+shutil.rmtree(base / '.github')
+shutil.rmtree(base / 'misc')
+shutil.rmtree(base / 'test/cases')
```
eli-schwartz, line 13: You can use gitattributes export-ignore here, but Meson only makes use of that starting with Meson 0.60...

meson.build
jaharkes: As far as I understood the meson documentation, there is a subtle difference.
The comment for the disabler argument on find_program states

> If `true` and the program couldn't be found, return a [disabler](https://mesonbuild.com/Reference-manual_returned_disabler.html) object

So specifying the `disabler : true` argument means that documentation would still be built as long as doxygen is found, while the original if else solution disables the documentation build when the doc option is set to disabled.

It might still be useful to add disabler:true to the find_program line either way so that we disable documentation build when required is false.

jaharkes: Ah, reading more, it looks like you may be right and the trick is that when the 'required' argument is disable, it will not try to look for the doxygen binary and as a result fail to find it and return the disabler.

So yes, that change LGTM

eli-schwartz: Indeed, feature options with a disabled state are IMHO one of the coolest things Meson has, because it allows you to concisely express things that a boolean simply cannot -- and part of that is that feature options Do The Right Thing™ without needing bulky if/else constructs to open-code what "disabled" means.

It will even give you a pretty log message: "Program doxygen skipped: feature **doxygen** disabled".

scripts/dist.py
jaharkes: considering that we're already constraining the environment where a source distributions is built with a specific version of doxygen it might just be worth the extra simplification to also assume meson >=0.60.

@bgilbert either way, the existing gitattributes is really autoconf specific, so either it can be removed completely as part of this patch, or should be updated to add export-ignore for those paths that should not be part of the source distribution.

bgilbert: Switched to gitattributes, and added a release checklist item to confirm Meson >= 0.60.

bgilbert:
Updated for comments.  @eli-schwartz, thanks for the review!

PR #431, opened 2023-03-15
DICOM vendor driver using libdicom

jcupitt:
This is a WSI DICOM loader for openslide using libdicom.

It can correctly load TILED_FULL dicom images and includes support for label and overview images. The 3D Histech demo DICOMs work well.

You give it the filename of one of the DICOMs in the pyramid you want to open, it then searches the enclosing directory and finds the other pyramid levels and any additional images with a matching slide id.

Still TODO in this PR:

- [x] I added libdicom to the ubuntu CI build, but I've not been able to get the windows build working yet
- [x] a test image (perhaps the 3dhistech demo DICOM)
- [x] enable fedora, macos, win test builds
- [x] libdicom should be an optional dependency

Other features (eg. sparse tile support, more metadata, etc) will be added in subsequent PRs.

Signed-off-by: John Cupitt <jcupitt@gmail.com>


fedorov:
@malaterre would you be interested to review this, given your authorship of #157?

jcupitt:
Here's a video I made of this PR in action:

bgilbert:
Thanks for the PR!  I think it's okay to introduce support for DICOM and then broaden it in subsequent PRs, as long as each PR: is believed to be solid, passes tests on all platforms, errors out if it encounters a slide file it can't handle, and ideally uses libdicom API that is believed to be stable.  Since libdicom hasn't had a stable release yet, we should make it an optional dependency at first, with the intent to switch to a mandatory dependency before the next OpenSlide release.

Alternatively, development could just continue in this PR.  If we need a middle ground, I'd be willing to consider merging incremental PRs to a separate branch in this repo, but would need to think about that a bit.

As to Windows support, we'd need a PR to [openslide-winbuild](https://github.com/openslide/openslide-winbuild) to add libdicom.  I understand that it'd need to point to a Git commit for now, but that should be okay, since the resulting DLL would initially be ignored except in CI and nightly builds.

We'll also need some CC0-licensed test files for our test suite.  They can be submitted [here](https://docs.google.com/forms/d/e/1FAIpQLSf4RsqIEVqB_FlY0mJtmi9fhHNSqc_59e4uHIoAgtkbtplH0w/viewform).  For us to use the slide, the answer to the last question on the form must give us permission to redistribute the file.  Since DICOM is a bit of a special case, we may be able to use data under a non-CC0 license, so long as the license is explicit and permissive; let me know if that approach will be needed.

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,943 @@
...
+  for (int i = 0; i < types->n_types; i++) {
+    bool found_difference = false;
+
+    for (int j = 0; j < 4; j++) {
+      const char *type;
+      if (!get_tag_str(f->metadata, "ImageType", j, &type)) {
+        return false;
+      }
+
+      if (strcmp(type, types->types[i][j]) != 0) {
```
bgilbert, line 460: OpenSlide usually uses `g_str_equal()` for clarity.

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,943 @@
...
+  // layer and pick that as the slide image we are opening
+  const struct level *largest = NULL;
+  g_hash_table_foreach(level_hash, find_largest, &largest);
+  if (!largest) {
+    return false;
+  }
+
+  const char *slide_id;
+  if (!get_tag_str(largest->file->metadata, 
+                   "SeriesInstanceUID",
```
bgilbert, line 856: Let's define all of these tags as `static const char[]` at the top of the file.

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,943 @@
...
+  guint32 *tiledata = _openslide_cache_get(osr->cache,
+                                            level, tile_col, tile_row,
+                                            &cache_entry);
+  if (!tiledata) {
+    g_auto(_openslide_slice) box =
+      _openslide_slice_alloc(tile_size * tile_size * 4);
+    guint32 frame_number = 1 + tile_col + l->tiles_across * tile_row;
+    if (frame_number < 1 || frame_number > l->num_frames) {
+      g_set_error(err, OPENSLIDE_ERROR, OPENSLIDE_ERROR_FAILED,
+                  "Frame number out of range");
```
bgilbert, line 350: Let's report the value and expected max.

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,943 @@
...
+  return dcm_element_get_value_string(NULL, element, index, result);
+}
+
+static struct dicom_file *dicom_file_new(char *filename, GError **err) {
+  struct dicom_file *f = g_slice_new0(struct dicom_file);
+  f->filename = g_strdup(filename);
+
+  // should be able to open as a DICOM
+  DcmError *dcm_error = NULL;
+  f->filehandle = dcm_filehandle_create_from_file(&dcm_error, f->filename);
```
bgilbert, line 276: We'll need to indirect libdicom's file accesses to go through OpenSlide's VFS.  Note that (for better or for worse) OpenSlide generally does not hold `FILE` pointers open between calls into the OpenSlide API.

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,943 @@
...
+                        const char *keyword, 
+                        int index, 
+                        const char **result) {
+  uint32_t tag = dcm_dict_tag_from_keyword(keyword);
+  DcmElement *element = dcm_dataset_get(NULL, dataset, tag);
+  return dcm_element_get_value_string(NULL, element, index, result);
+}
+
+static struct dicom_file *dicom_file_new(char *filename, GError **err) {
+  struct dicom_file *f = g_slice_new0(struct dicom_file);
```
bgilbert, line 271: `g_autoptr()`

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,943 @@
...
+  // reject TIFFs
+  if (tl) {
+    g_set_error(err, OPENSLIDE_ERROR, OPENSLIDE_ERROR_FAILED,
+                "Is a TIFF file");
+    return false;
+  }
+
+  // verify existence
+  GError *tmp_err = NULL;
+  if (!_openslide_fexists(filename, &tmp_err)) {
```
bgilbert, line 210: Is this necessary, or can we just let the open call fail if the file is missing?

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,943 @@
...
+  associated_type_strings, 
+  sizeof(associated_type_strings) / sizeof(associated_type_strings[0])
+};
+
+G_DEFINE_AUTOPTR_CLEANUP_FUNC(DcmFilehandle, dcm_filehandle_destroy)
+G_DEFINE_AUTOPTR_CLEANUP_FUNC(DcmDataSet, dcm_dataset_destroy)
+
+static void set_gerror_from_dcm_error(GError **err, DcmError **dcm_error)
+{
+  char *msg = g_strdup_printf("libdicom %s: %s - %s",
```
bgilbert, line 143: `g_autofree`

bgilbert (commented):
Initial review for structure and OpenSlide conventions.  I haven't yet read the logic in detail.

src/openslide.c
```diff
@@ -42,6 +42,7 @@ static const char * const EMPTY_STRING_ARRAY[] = { NULL };
 static const struct _openslide_format *formats[] = {
   &_openslide_format_synthetic,
   &_openslide_format_mirax,
+  &_openslide_format_dicom,
```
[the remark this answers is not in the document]
jcupitt: I can't think of a good ordering reason. It should be fairly quick at rejecting non-DICOMs and I think it's very unlikely (though technically possible I guess) to detect a non-DICOM as a DICOM.

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,943 @@
...
+
+#include <glib.h>
+#include <glib-object.h>
+#include <gio/gio.h>
+#include <string.h>
+#include <errno.h>
+
+#include <dicom/dicom.h>
+
+#define FREEF(F, V) { \
```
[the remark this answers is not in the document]
jcupitt: OK, removed.

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,943 @@
...
+    if (V) { \
+        (F)((V)); \
+        (V) = NULL; \
+    } \
+}
+
+/* The SOP we check for.
+ */
+#define VLWholeSlideMicroscopyImageStorage "1.2.840.10008.5.1.4.1.1.77.1.6"
```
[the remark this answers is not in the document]
jcupitt: `VLWholeSlideMicroscopyImageStorage` is the name of a DICOM constant, so I kept it. 

We could rename this as `DICOM_WSI_UID`, perhaps?

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,943 @@
...
+
+/*
+#define DEBUG
+ */
+
+enum image_format {
```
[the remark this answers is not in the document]
jcupitt: Yes, it's more of a placeholder. At least plain RGB image arrays are going in RSN (Leica use them for label images, bizarrely) it it's going to used very soon.

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,943 @@
...
+
+struct dicom_ops_data {
+  char *dirname;
+  int tile_size;
+};
+
+struct dicom_file {
+  char *filename;
+
+  GMutex lock;
```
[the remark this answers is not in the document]
jcupitt: Yes, it's regrettable. libdicom does a `seek()` then a `read()` to get a frame from the file, so if two threads use `get_frame` together, they can clash. It's just like TIFF read.

We'd need to add pthreads and all that machinery to libdicom to fix this, and I'd like to leave that library with no dependencies, if possible.

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,943 @@
...
+  data->dirname = g_strdup(dirname);
+  gsize count;
+
+  /* g_ptr_array_steal() is only available in recent glib ... hack around 
+   * this for older versions.
+   */
+#if GLIB_CHECK_VERSION(2, 64, 0)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+  osr->levels = (struct _openslide_level **) g_ptr_array_steal(level_array,
```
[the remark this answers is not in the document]
jcupitt: Oh haha I missed that. Thanks!

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,943 @@
...
+
+  const char *slide_id;
+  if (!get_tag_str(largest->file->metadata, 
+                   "SeriesInstanceUID", 
+                   0, 
+                   &slide_id)) {
+    return false;
+  }
+
+  // throw away all files which don't have this slide_id
```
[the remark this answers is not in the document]
jcupitt: I agree, I'll change this to take the slide id from the original file, not the largest level.

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,943 @@
...
+
+static int find_associated(gpointer key G_GNUC_UNUSED, 
+                            gpointer value, 
+                            gpointer user_data) {
+  struct dicom_file *f = (struct dicom_file *) value;
+  openslide_t *osr = (openslide_t *) user_data; 
+
+  struct associated *a = associated_new(f);
+  if (a) {
+    // FIXME ... this will use LABEL and OVERVIEW ... is this OK?
```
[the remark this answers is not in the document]
jcupitt: Ah I missed that bit of the docs, thanks!

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,943 @@
...
+
+  return a;
+}
+
+static struct level *level_new(struct dicom_file *f) {
+  if (!read_whole_file(f)) { 
+    return NULL;
+  }
+
+  struct level *l = g_slice_new0(struct level);
```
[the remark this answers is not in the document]
jcupitt: This is a constructor, so I think we need this to live outside the scope of this function. 

I allocated the `level` before I was certain construction was going to succeed since I wanted somewhere to note the file properties it has to check. I could read the properties to locals instead if that would be preferred.

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,943 @@
...
+    g_hash_table_new_full(g_str_hash, 
+                          g_str_equal, 
+                          g_free,
+                          (GDestroyNotify) dicom_file_destroy);
+
+  // open all DICOM files that look like parts of a slide image and 
+  // get the file metadata
+  const char *name;
+  while ((name = g_dir_read_name(dir))) {
+    char *filename = g_build_path("/", dirname, name, NULL);
```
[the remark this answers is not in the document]
jcupitt: `filename` is stolen by the hash table, so I didn't want to use autoptr here. I could copy the string instead of copying the pointer if that would be more openslidey.

src/openslide-vendor-dicom.c
jcupitt: OK, I'll change this.

src/openslide-vendor-dicom.c
jcupitt: OK, done! I didn't use a prefix on the var names, hopefully that's OK.

src/openslide-vendor-dicom.c
jcupitt: This is a constructor, so I think autoptr won't work, will it?

src/openslide-vendor-dicom.c
jcupitt: OK, gone!

src/openslide-vendor-dicom.c
jcupitt: Ooops, missed that.

src/openslide-vendor-dicom.c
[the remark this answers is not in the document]
jcupitt: ... I added:

```C
static const char VLWholeSlideMicroscopyImageStorage[] =
  "1.2.840.10008.5.1.4.1.1.77.1.6";
```

at the top of the file, along with the other DICOM names we use.

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,943 @@
...
+}
+
+static void set_downsample(gpointer key G_GNUC_UNUSED, 
+                           gpointer value, 
+                           gpointer user_data) {
+  struct level *l = (struct level *) value;
+  const struct level *largest = (struct level *) user_data;
+
+  // need to compute downsample now we have largest
+  int downsample = largest->image_width / l->image_width;
```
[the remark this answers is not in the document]
jcupitt: DICOMs are always strict powers of two (phew!), but you're right, the downsampling code in `openslide.c` will work here and save us some lines, I'll change it.

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,943 @@
...
+    dcm_error_summary(*dcm_error),
+    dcm_error_message(*dcm_error));
+  g_set_error(err, OPENSLIDE_ERROR, OPENSLIDE_ERROR_FAILED, "%s", msg);
+  g_free(msg);
+  dcm_error_clear(dcm_error);
+}
+
+#ifdef DEBUG
+static void print_file(struct dicom_file *f) {
+  printf("file:\n" );
```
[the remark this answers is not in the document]
jcupitt: It's very verbose right now (I like reading logs), I was planning to remove most of this stuff later in development when there's less basic machinery that needs verifying. Would that be OK?

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,943 @@
...
+}
+
+static bool dicom_open(openslide_t *osr, 
+                       const char *filename,
+                       struct _openslide_tifflike *tl G_GNUC_UNUSED,
+                       struct _openslide_hash *quickhash1, 
+                       GError **err) {
+  g_autofree char *dirname = g_path_get_dirname(filename);
+
+  g_autoptr(GDir) dir = g_dir_open(dirname, 0, err);
```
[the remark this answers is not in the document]
jcupitt: Ah I didn't realize openslide had a VFS layer. I'll do a separate commit for openslide VFS integration.

src/openslide-vendor-dicom.c
jcupitt: Implemented with https://github.com/openslide/openslide/pull/431/commits/3d2f50d72775d5e5ab0360cd2edff99922c56ce9

src/openslide-vendor-dicom.c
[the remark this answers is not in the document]
bgilbert: Separate commit in the same PR should be fine.  And yes, I'd favor the glib directory functions rather than dealing with platform quirks ourselves.

src/openslide-vendor-dicom.c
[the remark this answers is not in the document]
bgilbert: Yup, exactly.  I generally don't bother with `g_autoptr` when we're just going to infallibly set some fields and return, but if there are error paths we generally avoid manual frees.

src/openslide-vendor-dicom.c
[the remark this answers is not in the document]
bgilbert: I'd be inclined to insert `g_steal_pointer(filename)` into the hash table, but inserting `g_strdup(filename)` would also be okay.  The idea is that a use-after-free in the happy path is easier to notice than a leak (or double free) in an error path.

src/openslide-vendor-dicom.c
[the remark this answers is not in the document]
bgilbert: Removing them later is okay.  The codebase does already have a bunch of commented-out `g_debug()` calls in case they need to be re-enabled later (not my first choice of approach, but I inherited it), so I think it's okay to leave some of the debug prints in the codebase long-term.  To reduce the code verbosity, though, how about something like:

```c
#if 1
#define debug(...) g_debug(__VA_ARGS__)
#else
#define debug(...)
#endif
```

src/openslide-vendor-dicom.c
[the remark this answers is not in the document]
jcupitt: OK, done!

src/openslide-vendor-dicom.c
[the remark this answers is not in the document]
jcupitt: OK, done!

src/openslide-vendor-dicom.c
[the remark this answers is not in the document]
bgilbert: Well, if it were me, I'd just call the debug functions unconditionally, and live with the small perf hit of iterating over some stuff only to call a debug function that's going to be compiled out.  As you say, this isn't long-term code.

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,961 @@
...
+static bool paint_region(openslide_t *osr, 
+                         cairo_t *cr,
+                         int64_t x, 
+                         int64_t y,
+                         struct _openslide_level *level,
+                         int32_t w, 
+                         int32_t h,
+                         GError **err) {
+  struct dicom_ops_data *data = osr->data;
+  Level *l = (Level *) level;
```
bgilbert, line 486: Please use typedefs in accordance with the [style guide](https://openslide.org/docs/newformat/#memory-allocation) (i.e., minimally).

jcupitt: OK, done (I hope).

src/openslide-vendor-dicom.c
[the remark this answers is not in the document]
jcupitt: OK, I swapped it to just use `g_debug()` everywhere. It's much too chatty, but as you say, we can remove this before release.

jcupitt:
For a test image, how about:

ftp://medical.nema.org/MEDICAL/Dicom/DataSets/WG26/Hackathon_2021/3DHISTECH/WithMetadata_20210527/3DHISTECH%5ESmoothMuscle%20%5BID_20210527084705%5D

This is a test DICOM released by 3D Histech in 2021 for the WG26 (the working group in DICOM responsible for the WSI standard) hackathon and available on the nema (the public face of WG26) public ftp. 

It works well with this PR. I can zip and upload it to the openslide slide submissions page if that would be helpful.

jcupitt:
I found this:

https://github.com/DICOMWG26AnnotationAHG/hackathon#data-use

> Any data participants contribute to the Hackathon (e.g., store in one of the provided image management systems) shall be made available to anyone under the [CCO](https://creativecommons.org/share-your-work/public-domain/cc0/) license. The data will subsequently be available to anyone for free and can be reused without restrictions, including for commercial purposes.

Is that sufficient?

fedorov:
@bgilbert we have a lot of slide microscopy DICOM data in IDC, which is covered by CC-BY. If CC0 is a must (and I am not sure I have ever seen DICOM data explicitly shared under CC0), we will need to look in other places. I am curious, was there any specific reason not to accept CC-BY?

@dclunie are you aware of any DICOM SM samples covered by CC0 license?

hackermd:
@fedorov @dclunie we required all files for recent DICOM WG-26 Connectathons and Hackathons to be shared under a COO license.

bgilbert:
@jcupitt Yes, that should be sufficient, thanks!  No need to fill out the form in that case.

@fedorov We chose CC0 to minimize restrictions on reuse of the data.  Since the goal is to capture the mechanical output of scanner software rather than any creative content in the images or metadata, protecting authorship rights wasn't a priority.  We do credit contributors ([example](https://openslide.cs.cmu.edu/download/openslide-testdata/Leica/)), we just don't require it via the license.

fedorov:
> @fedorov We chose CC0 to minimize restrictions on reuse of the data. Since the goal is to capture the mechanical output of scanner software rather than any creative content in the images or metadata, protecting authorship rights wasn't a priority. 

I understand you want to minimize restrictions. But minimizing restrictions and bringing those to 0 is not the same, and I personally think requirement to CC0 exclusively is potentially problematic.

The issue is that in many (most?) cases whole slide imaging data from human specimens is collected as part of a clinical research study, and many steps need to take place before (get the funding to do all of the following, design the study, get IRB protocols in place, consent patients, establish infrastructure, allocate personnel, etc) and after (work with the data collection center, define de-identification procedures, implement de-identification, perform QA, establish archival and storage infrastructure, support query and retrieval, etc) capturing the mechanical output of scanner software. And also, someone would need to provision the scanning equipment, which most often than not is not used in the clinic. Further, when it comes to DICOM, most often scanners don't produce DICOM output, but save images in vendor-specific format, so someone would need to work on harmonization and conversion. 

Those "before", and "after" steps are, unfortunately, most often overlooked, but take HUGE effort. Most often than not, image data is not shared at all, in particular because those who do invest effort into sharing do not get back anything to help their careers - they often do it just because they believe it is the right thing to do for the science. CC-BY constraints are absolutely minimal, and reducing the process of getting the image to someone pushing the button on the scanner console diminishes the effort of many people contributing to collection and sharing of this data. It is truly unfortunate that even such minimal requirement is considered to be burdensome, given the huge effort that goes into the data collection and sharing, and maintenance of the resources hosting those datasets.

jaharkes:
@fedorov  The CC0 choice is for any slide data used in the openslide regression test set which is used to validate that whenever code is changed it doesn't accidentially break known supported slide formats. A simple "Hi Guys!" written on a slide with a sharpie will probably work just as well as anything, and might even compress better anyway.

There is no need for human specimens, studies, IRBs, patient consent, deindentification, and, once added to the test set, even archiving and storage should not have to be your concern.

Whatever slide data you process with Openslide for your own use cases can be licensed it any way you want to, there are no restrictions there.

.github/workflows/c.yaml
```diff
@@ -252,7 +260,8 @@ jobs:
   windows_build:
     name: Windows build
     needs: windows_setup
-    uses: openslide/openslide-winbuild/.github/workflows/windows.yml@main
+    uses: jcupitt/openslide-winbuild/.github/workflows/windows.yml@add-libdicom-prototype
```
bgilbert, line 263: This is causing the workflow not to start:

```
error parsing called workflow
".github/workflows/c.yaml"
-> "jcupitt/openslide-winbuild/.github/workflows/windows.yml@add-libdicom-prototype"
: workflow was not found. See https://docs.github.com/actions/learn-github-actions/reusing-workflows#access-to-reusable-workflows for more information.
```

.github/workflows/c.yaml
```diff
@@ -143,7 +137,21 @@ jobs:
...
+        if ! git clone -b $libdicom_branch $libdicom_url; then
+          exit 1
+        fi
+        cd libdicom
+        if ! meson setup builddir; then
+          cat builddir/meson-logs/meson-log.txt
+          exit 1
+        fi
+        ninja -C builddir
+        $sudo ninja -C builddir install
```
bgilbert, line 154: Aside from this block, I think all of the changes in this file are superfluous?  We'll need macOS CI, and we need to support the Meson version that ships natively with each distro.  If there are changes in this file that you feel are necessary, let's discuss.

Nit: indentation should be four spaces.

jcupitt: Yes, it's still a draft while I learn the meson wrap system. I'll clean it up and squash it before making it final.

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,932 @@
...
+      }
+
+      if (!g_str_equal(type, types->types[i][j])) {
+        found_difference = true;
+        break;
+      }
+    }
+
+    if (!found_difference) {
+      match = i;
```
bgilbert, line 518: We could just `return true` here and drop `match`.

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,932 @@
...
+                                          l->tile_h,
+                                          read_tile);
+}
+
+static void add_properties(openslide_t *osr, const struct dicom_level *l) {
+  // why not
+  double mmpp = l->image_width;
+  g_hash_table_insert(osr->properties,
+                      g_strdup(OPENSLIDE_PROPERTY_NAME_MPP_X),
+                      _openslide_format_double(mmpp * 1000));
```
bgilbert, line 779: Wait, so we're saying the number of microns per pixel is exactly 1000x the number of pixels in X?  That doesn't seem right.

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,932 @@
...
+}
+
+static void file_hash(const struct dicom_file *f,
+                      struct _openslide_hash *quickhash1) {
+  uint32_t n = dcm_dataset_count(f->metadata);
+  for(uint32_t i = 0; i < n; i++) {
+    DcmElement *element = dcm_dataset_get(NULL, f->metadata, i);
+    if (element) {
+      uint32_t tag = dcm_element_get_tag(element);
+      _openslide_hash_data(quickhash1, &tag, sizeof(tag));
```
bgilbert, line 792: Could you spell out exactly what this is doing?  We need to be careful about the definition of the quickhash since it needs to be stable, different for every slide, and fast.  We usually put the quickhash definition in a comment at the top of the source file.

Note that we often hash in the image data from the smallest slide level.

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,932 @@
...
+  // build ops data
+  struct dicom_ops_data *data = g_slice_new0(struct dicom_ops_data);
+  data->dirname = g_strdup(dirname);
+
+  guint count = level_array->len;
+  // we need to use g_ptr_array_free() to steal the array for openslide, but
+  // level_array is an autoptr ... add an extra ref
+  g_ptr_array_ref(level_array);
+  osr->levels = (struct _openslide_level **) g_ptr_array_free(level_array, 
+                                                              false);
```
bgilbert, line 906: This should just be `g_ptr_array_free(g_steal_pointer(&level_array), false)`.

src/openslide-vendor-dicom.c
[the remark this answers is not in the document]
bgilbert: This is okay for now, but I think the right long-term solution is to generalize tiffcache into a generic handle cache and use it here.  I might give that a shot at some point.  It would solve two problems:

- OpenSlide generally avoids holding file handles open between calls, and the driver is currently doing that.
- We could get rid of the lock and let multiple threads decode in parallel by using multiple `DcmFilehandle`s.

bgilbert (commented):
Detailed review; mostly found small things.  The overall structure looks good except for the hash table callbacks invoked from `dicom_open`, which shouldn't be too hard to refactor.

bgilbert:
Directly pushed some small cleanups rather than posting review comments asking you to copyedit.

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,914 @@
...
+static const char SeriesInstanceUID[] = "SeriesInstanceUID";
+static const char TotalPixelMatrixColumns[] = "TotalPixelMatrixColumns";
+static const char TotalPixelMatrixRows[] = "TotalPixelMatrixRows";
+static const char Columns[] = "Columns";
+static const char Rows[] = "Rows";
+
+G_DEFINE_AUTOPTR_CLEANUP_FUNC(DcmFilehandle, dcm_filehandle_destroy)
+G_DEFINE_AUTOPTR_CLEANUP_FUNC(DcmDataSet, dcm_dataset_destroy)
+
+static void set_gerror_from_dcm_error(GError **err, DcmError *dcm_error) {
```
bgilbert, line 131: For consistency with elsewhere in the library, let's use `propagate` as the verb in these function names.

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,932 @@
...
+  _openslide_fclose(file);
+  return 0;
+}
+
+static int64_t dicom_openslide_vfs_read(DcmError **dcm_error G_GNUC_UNUSED, 
+                                        void *data, 
+                                        char *buffer, 
+                                        int64_t length) {
+  struct _openslide_file *file = (struct _openslide_file *) data;
+  // openslide VFS has no error return for read()
```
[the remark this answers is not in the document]
jcupitt: I don't think this is possible -- stdio `fread()` doesn't distinguish between EOF and read error, you have to use `feof()` and `ferror()` to tell them apart, and they are not exposed by the openslide VFS.

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,932 @@
...
+  }
+
+  // set this once we know we will succeed ... this passes ownership to the
+  // level
+  l->file = f;
+  l->num_frames = dcm_bot_get_num_frames(l->file->bot);
+  l->base.w = l->image_width;
+  l->base.h = l->image_height;
+  l->base.tile_w = l->tile_w;
+  l->base.tile_h = l->tile_h;
```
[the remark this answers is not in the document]
jcupitt: OK, done.

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,932 @@
...
+    return NULL;
+  }
+
+  const char *slide_id;
+  if (!get_tag_str(start->metadata, SeriesInstanceUID, 0, &slide_id)) {
+    return false;
+  }
+
+  // throw away all files which don't have this slide_id
+  g_hash_table_foreach_remove(dicom_file_hash,
```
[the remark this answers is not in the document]
jcupitt: Good idea, done.

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,932 @@
...
+  if (l->tile_w != l->tile_h) {
+    return NULL;
+  }
+
+  // set this once we know we will succeed ... this passes ownership to the
+  // level
+  l->file = f;
+  l->num_frames = dcm_bot_get_num_frames(l->file->bot);
+  l->base.w = l->image_width;
+  l->base.h = l->image_height;
```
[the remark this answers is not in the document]
jcupitt: OK, done.

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,932 @@
...
+  DcmError *dcm_error = NULL;
+  f->meta = dcm_filehandle_read_file_meta(&dcm_error, 
+                                          f->filehandle);
+  if (!f->meta) {
+    set_gerror_from_dcm_error(err, &dcm_error);
+    return NULL;
+  }
+
+  const char *sop;
+  get_tag_str(f->meta, "MediaStorageSOPClassUID", 0, &sop);
```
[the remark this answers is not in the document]
jcupitt: I'm expecting to have to read more stuff out of meta in phases 2 and 3, so I'd like to keep this for now.

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,932 @@
...
+
+#include <dicom/dicom.h>
+
+enum image_format {
+  FORMAT_UNKNOWN,
+  FORMAT_JPEG,
+};
+
+struct dicom_ops_data {
+  char *dirname;
```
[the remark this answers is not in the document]
jcupitt: OK, removed.

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,932 @@
...
+  int64_t tile_size;
```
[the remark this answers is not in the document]
jcupitt: OK, removed.

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,932 @@
...
+  // cache
+  g_autoptr(_openslide_cache_entry) cache_entry = NULL;
+  guint32 *tiledata = _openslide_cache_get(osr->cache,
+                                            level, tile_col, tile_row,
+                                            &cache_entry);
+  if (!tiledata) {
+    g_auto(_openslide_slice) box =
+      _openslide_slice_alloc(tile_size * tile_size * 4);
+    guint32 frame_number = 1 + tile_col + l->tiles_across * tile_row;
+    if (frame_number < 1 || frame_number > l->num_frames) {
```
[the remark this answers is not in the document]
jcupitt: OK, removed if this is guaranteed.

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,932 @@
...
+static const char TotalPixelMatrixRows[] = "TotalPixelMatrixRows";
+static const char Columns[] = "Columns";
+static const char Rows[] = "Rows";
+
+G_DEFINE_AUTOPTR_CLEANUP_FUNC(DcmFilehandle, dcm_filehandle_destroy)
+G_DEFINE_AUTOPTR_CLEANUP_FUNC(DcmDataSet, dcm_dataset_destroy)
+
+static void set_gerror_from_dcm_error(GError **err, DcmError **dcm_error)
+{
+  g_autofree char *msg = g_strdup_printf("libdicom %s: %s - %s",
```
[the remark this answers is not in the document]
jcupitt: OK, removed.

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,932 @@
...
+}
+
+G_DEFINE_AUTOPTR_CLEANUP_FUNC(dicom_file, dicom_file_destroy)
+
+static bool get_tag_int(DcmDataSet *dataset, 
+                        const char *keyword, 
+                        int64_t *result) {
+  uint32_t tag = dcm_dict_tag_from_keyword(keyword);
+  DcmElement *element = dcm_dataset_get(NULL, dataset, tag);
+  return dcm_element_get_value_integer(NULL, element, 0, result);
```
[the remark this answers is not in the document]
jcupitt: OK, guard added.

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,932 @@
...
+  f->meta = dcm_filehandle_read_file_meta(&dcm_error, 
+                                          f->filehandle);
+  if (!f->meta) {
+    set_gerror_from_dcm_error(err, &dcm_error);
+    return NULL;
+  }
+
+  const char *sop;
+  get_tag_str(f->meta, "MediaStorageSOPClassUID", 0, &sop);
+  if (!g_str_equal(sop, VLWholeSlideMicroscopyImageStorage)) {
```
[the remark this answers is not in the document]
jcupitt: OK, guard added.

src/openslide-vendor-dicom.c
jcupitt: OK, changed.

src/openslide-vendor-dicom.c
jcupitt: It's just a placeholder until we add full metadata support (coming in phase 2).

src/openslide-vendor-dicom.c
jcupitt: This is a placeholder. I'm planning to fix this up in phase 2 when we get more complete metadata support.

src/openslide-vendor-dicom.c
jcupitt: OK, changed.

src/openslide-vendor-dicom.c
jcupitt: OK, I swapped the names.

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,932 @@
...
+    return false;
+  }
+
+  // throw away all files which don't have this slide_id
+  g_hash_table_foreach_remove(dicom_file_hash, 
+                              remove_bad_slide_id, 
+                              (char *) slide_id);
+
+  // pull out the subset of DICOM files that look like pyramid levels
+  g_autoptr(GHashTable) level_hash =
```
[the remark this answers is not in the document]
jcupitt: That's a bit awkward since the passed-in filename can be either an associated image or a level or perhaps neither but part of the same slide. You end up needing some confusing extra ifs.

I trimmed it down to just a hash for the matching slide ids and a ptrarray for the levels, is that enough?

.github/workflows/c.yaml
jcupitt: ooop, thanks! I thought `main` was a branch name.

src/openslide-vendor-dicom.c
bgilbert: Let's not intentionally add incorrect metadata.  We can drop this for now and re-add in phase 2.

src/openslide-vendor-dicom.c
bgilbert: We shouldn't change an existing quickhash once it's been introduced.  For now, we can call `_openslide_hash_disable()`.

src/openslide-vendor-dicom.c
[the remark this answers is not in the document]
bgilbert: I think I see what you mean about the extra ifs, but that can be solved with an "add file" function.  I think that's less confusing than calling out to separate functions when iterating.

(And actually, we don't need a separate array for associated images, since we can insert directly into `osr->associated_images` as we go.)

I'm imagining the logic would look roughly like this (high-level pseudocode, not meant literally):

```
func add_file(osr, levels, file) {
  if is_level(file) {
    levels += file
  } else if is_associated(file) {
    osr.associated_images[associated_image_name(file)] = file
  }
}

func dicom_open(osr, filename) {
  base_file = open(filename)
  slide_id = slide_id_of(base_file)
  levels = []
  add_file(osr, levels, base_file)
  foreach file in walkdir(dirname(filename)) {
    // if this check is too hard, we could skip the add_file() above,
    // and maybe inline add_file() below
    if file == base_file {
      continue
    }
    cur_slide_id = slide_id_of(file)
    if cur_slide_id != slide_id {
      continue
    }
    add_file(osr, levels, file)
  }
  sort(levels)
  ...
}
```

src/openslide-vendor-dicom.c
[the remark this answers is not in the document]
bgilbert: Well, the point of the handle cache is that we'd only allocate additional `DcmFilehandle`s when we actually need them.

.github/workflows/c.yaml
```diff
@@ -252,7 +268,8 @@ jobs:
   windows_build:
     name: Windows build
     needs: windows_setup
-    uses: openslide/openslide-winbuild/.github/workflows/windows.yml@main
+    uses: jcupitt/openslide-winbuild/.github/workflows/windows.yml@main
+    # uses: openslide/openslide-winbuild/.github/workflows/windows.yml@main
```
bgilbert, line 272: Any reason we can't use the upstream one?

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,840 @@
...
+#include <glib.h>
+
+#include <dicom/dicom.h>
+
+enum image_format {
+  FORMAT_UNKNOWN,
+  FORMAT_JPEG,
+};
+
+struct dicom_ops_data {
```
bgilbert, line 41: We can drop this.  :slightly_smiling_face:

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,840 @@
...
+  DcmDataSet *meta;
+  DcmBOT *bot;
+};
+
+struct dicom_level {
+  struct _openslide_level base;
+  struct _openslide_grid *grid;
+
+  enum image_format image_format;
+  guint32 num_frames;
```
bgilbert, line 59: Unused.

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,840 @@
...
+
+  g_debug("read_tile: tile_col = %" PRIu64 ", tile_row = %" PRIu64,
+          tile_col,
+          tile_row);
+  g_debug("read_tile level:");
+  print_level(l);
+
+  // cache
+  g_autoptr(_openslide_cache_entry) cache_entry = NULL;
+  guint32 *tiledata = _openslide_cache_get(osr->cache,
```
bgilbert, line 382:
```suggestion
  uint32_t *tiledata = _openslide_cache_get(osr->cache,
```

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,840 @@
...
+                                            level, tile_col, tile_row,
+                                            &cache_entry);
+  if (!tiledata) {
+    g_auto(_openslide_slice) box =
+      _openslide_slice_alloc(tile_size * tile_size * 4);
+    guint32 frame_number = 1 + tile_col + l->tiles_across * tile_row;
```
bgilbert, line 388:
```suggestion
    uint32_t frame_number = 1 + tile_col + l->tiles_across * tile_row;
```

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,840 @@
...
+static const char *overview_types[4] = {
+  "ORIGINAL", "PRIMARY", "OVERVIEW", "NONE",
+};
+static const char **associated_type_strings[] = {
+  label_types,
+  overview_types
+};
+static const struct allowed_types associated_types = {
+  associated_type_strings,
+  sizeof(associated_type_strings) / sizeof(associated_type_strings[0])
```
bgilbert, line 109: `G_N_ELEMENTS`

src/openslide-vendor-dicom.c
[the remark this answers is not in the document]
bgilbert: Hmm, I'm not seeing it?

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,840 @@
...
+}
+
+static const struct _openslide_associated_image_ops dicom_associated_ops = {
+  .get_argb_data = associated_get_argb_data,
+  .destroy = openslide_associated_destroy
+};
+
+static bool read_whole_file(struct dicom_file *f) {
+  // try to read the rest of the dicom file
+  if (!f->metadata) {
```
bgilbert, line 539: Ideally we'd get to the point where this function is called at most once per `dicom_file` (or inlined into `dicom_open`) and we don't need the conditionals.

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,840 @@
...
+    !get_tag_int(f->metadata, Columns, &l->base.tile_w) ||
+    !get_tag_int(f->metadata, Rows, &l->base.tile_h)) {
+    return NULL;
+  }
+
+  // ImageType must be one of the combinations we accept
+  if (!is_type(f, &level_types)) {
+    return NULL;
+  }
+
```
bgilbert, line 613: Should we also check for JPEG, and for Z-stack layers other than the one we'll be selecting?  Fluorescence?  Any other parameters we don't support?

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,840 @@
...
+  // we only allow square tiles
+  if (l->base.tile_w != l->base.tile_h) {
+    return NULL;
```
bgilbert, line 616: Currently, when this function rejects a level, we just continue looking for other levels.  That makes sense when ignoring truly optional data (e.g. maybe an associated image we don't understand), but OpenSlide generally tries to reject slides it doesn't understand rather than silently producing incomplete data or garbage.  Should some of these tests be converted into hard failures?

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,840 @@
...
+    l->grid = _openslide_grid_create_simple(osr,
+					    l->tiles_across,
+					    l->tiles_down,
+					    l->base.tile_w,
+					    l->base.tile_h,
+					    read_tile);
+  }
+
+  // now sort levels by base.w to make the level array
+  g_ptr_array_sort(level_array, compare_level_width);
```
bgilbert, line 800: It's possible that we could end up with zero levels here.  We should detect that case and fail.

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,840 @@
...
+  struct associated *a = (struct associated *) _img;
+  associated_destroy(a);
+}
+
+static const struct _openslide_associated_image_ops dicom_associated_ops = {
+  .get_argb_data = associated_get_argb_data,
+  .destroy = openslide_associated_destroy
+};
+
+static bool read_whole_file(struct dicom_file *f) {
```
bgilbert, line 537: We should probably propagate errors to the caller, which can decide what to do with them (such as logging under a debug flag).  In general, functions should propagate any errors, unless they're a convenience function e.g. `get_tag_int` that has a reason not to.

bgilbert (commented):
Thanks for the fixes.  I'm afraid there are some new comments that I missed in the previous review pass.  I'm still getting my head around some of the implications of the code.

.github/workflows/c.yaml
bgilbert: Is the plan to ship a libdicom wrap in OpenSlide, or to have CI explicitly build libdicom?

Also, it looks like you're continuing to work on having CI upgrade Meson in the various distros.  CI should test what we actually need to support, which is the Meson version shipped by the distros.  What problem are you trying to solve there?

bgilbert:
Windows CI is failing because libdicom doesn't set up `version.h` to be found correctly when built as a subproject:

```
In file included from ../../meson/subprojects/openslide/src/openslide-vendor-dicom.c:34:
../../meson/subprojects/libdicom/include/dicom/dicom.h:5:10: fatal error: version.h: No such file or directory
    5 | #include "version.h"
      |          ^~~~~~~~~~~
compilation terminated.
```

.github/workflows/c.yaml
jcupitt: libdicom needs a more recent meson than the version shipped in at least centos8 and ubuntu 22.04.

jcupitt: ... I was thinking of including a libdicom wrap, yes.

src/openslide-vendor-dicom.c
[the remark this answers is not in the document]
jcupitt: Is it worth polishing this PR too much in this cycle? We're going to be going over this same code in at least two more PRs within a few weeks. A lot of this work now will have to be redone anyway.

src/openslide-vendor-dicom.c
jcupitt: It's just a placeholder :( it'll be replaced within a couple of weeks.

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,840 @@
...
+  struct dicom_file *start = dicom_file_new(filename, err);
+  if (!start) {
+    return false;
+  }
+  g_hash_table_insert(dicom_file_hash, g_strdup(filename), start);
+
+  const char *slide_id;
+  if (!read_whole_file(start) ||
+    !get_tag_str(start->metadata, SeriesInstanceUID, 0, &slide_id)) {
+    return false;
```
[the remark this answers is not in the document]
jcupitt: Ooops! Fixed.

.github/workflows/c.yaml
jcupitt: Now it's merged, sure. I'll flip it over.

src/openslide-vendor-dicom.c
jcupitt: OK, removed.

src/openslide-vendor-dicom.c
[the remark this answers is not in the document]
jcupitt: Ooop, I got my guards mixed up, sorry. Fixed now.

src/openslide-vendor-dicom.c
```diff
@@ -0,0 +1,840 @@
...
+  }
+
+  // open all DICOM files that look like parts of this slide image and
+  // get the file metadata
+  const char *name;
+  while ((name = _openslide_dir_next(dir))) {
+    g_autofree char *path = g_build_path("/", dirname, name, NULL);
+
+    // no need to open the start file again
+    if (g_str_equal(path, filename)) {
```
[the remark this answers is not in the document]
jcupitt: OK, changed to check basename.

src/openslide-vendor-dicom.c
jcupitt: I'm expecting to expand this -- it's another placeholder.

src/openslide-vendor-dicom.c
jcupitt: I think this is OK for now. We're just implementing visible light slides.

src/openslide-vendor-dicom.c
jcupitt: DICOM is extremely flexible and extensible, even more than tiff, which is a blessing and a curse. I think DICOM viewers are expected to display what they can.

src/openslide-vendor-dicom.c
jcupitt: OK, added an error.

src/openslide-vendor-dicom.c
jcupitt: We have to call `read_whole_file()` before we can get any of the main metadata, which we have to do decide if this file is one of the things we can handle. So I think this is more of a tester (like `get_tag`) than an action the loader is performing that can fail.

src/openslide-vendor-dicom.c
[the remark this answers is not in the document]
jcupitt: Unfortunately `DcmFilehandle`s are huge -- each one would parse the complete file again and allocate another copy of every element. It's not a filehandle in the windows sense -- it's a thing that represents the parsed state of DICOM File (in the dicom meaning).

src/openslide-vendor-dicom.c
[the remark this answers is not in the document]
jcupitt: Oooof, ok, reorganised as requested.

.github/workflows/c.yaml
bgilbert: https://github.com/ImagingDataCommons/libdicom/pull/47 reduces libdicom's minimum Meson version to older than OpenSlide's minimum.

jcupitt:
I've merged your commit to drop meson in libdicom to 0.5 into my fork, so openslide CI should also work with meson 0.5.

.github/workflows/c.yaml
bgilbert: And https://github.com/ImagingDataCommons/libdicom/pull/49 is needed to find uthash correctly on Ubuntu 20.04.

subprojects/libdicom.wrap
```diff
@@ -0,0 +1,6 @@
+[wrap-git]
+# NOTE: temporary fork for now
+# https://github.com/openslide/openslide-winbuild/issues/82
+url = https://github.com/jcupitt/libdicom.git
+revision = d7c08d4b89a8ea5e53511534be10c76826d66d7b
```
jcupitt, line 5: I suppose this should be updated to 2f3c8ff54ab7238f7aa4faba96ae41752d505769

rferreiroa:
Good morning,

i have a question regarding this feature. Whe this it is finished, openslide can manage dicom pathology images, or it is only for a specific vendor.

Do you know if OpenSeaDragon can render dicom images. (Because there is an example in openslide-python with this library)

thank you in advance and kind regards.

bgilbert:
The goal is to support DICOM WSI images from a variety of vendors.  OpenSeadragon can be used with an OpenSlide-based tile server as the frontend to any format supported by OpenSlide.

bgilbert (approved):

jcupitt:
> Whe this it is finished, openslide can manage dicom pathology images, or it is only for a specific vendor.

This is phase one of a project to implement DICOM for openslide. Support for other vendors is coming in phase 2, hopefully by June.

hackermd:
@jcupitt and @bgilbert that's a big milestone. Thank you for all your time and effort on making this happen. I am happy to see libdicom being used by openslide to read DICOM files.

jeffshek:
Representing on behalf of histowiz.com, we do have dicom files scanned directly from Pramana and Lecia GT450 scanners that we can send under the licenses you need. 

If you even want specific physical slides scanned for this project, we'll gladly do it for free.

Feel free to reach out to me at jeff@histowiz.com - we test through quite a bit of conditions (rescans, etc) so we have a lot of test files.

jcupitt:
Hi @jeffshek, that's great, thank you!

I've been testing with GT450 scans, so they ought to work well. I've not tried a Pramana. I've been using the slides in the nema archive.

What we really need is more testing -- people to build this thing (ie. openslide master), try it on their data, and report any errors they encounter in a way that we can reproduce. We were hoping to have this released by the end of July, but we've crashed into the summer break and it's now looking like September. So we have a few weeks for further testing before this goes fully public.

bgilbert:
@jeffshek Thank you very much!  I'd amend @jcupitt's comments a bit: we do need more manual testing, but we also need more test data.  New slides added to [openslide-testdata](https://openslide.cs.cmu.edu/download/openslide-testdata/) are generally required to use the [CC0 license](https://creativecommons.org/share-your-work/public-domain/cc0/), and since that's more permissive than most other slide sources, we often can't reuse existing samples in our automated tests.

We already have a few [samples](https://openslide.cs.cmu.edu/download/openslide-testdata/DICOM/), including a GT450 one, but more would be very welcome.  A Pramana sample would be great, as would samples of features we don't have in our test suite, such as Z-stacks, high bit depth, or new transfer syntaxes.  It sounds like you might have some other corner cases as well?

You can submit samples [here](https://docs.google.com/forms/d/e/1FAIpQLSf4RsqIEVqB_FlY0mJtmi9fhHNSqc_59e4uHIoAgtkbtplH0w/formResponse).  If you have some flexibility in selecting slides, smaller ones (several hundred MB rather than multiple GB) would be more convenient in tests.

Please let me know what I can do to help!

dclunie:
@bgilbert FYI re. z-stack DICOM samples, on [news:comp.protocols.dicom](https://groups.google.com/g/comp.protocols.dicom/c/j0YAu_Jtnq0)

taanil:
@bgilbert We have developed a new software to convert our scanner generated images into DICOM format. I have submitted sample images using Google Forms. The sample images are brightfield image series, with single channel and multiple focus planes.

Some initial discussion can be found here, https://groups.google.com/g/comp.protocols.dicom/c/j0YAu_Jtnq0

PR #439, opened 2023-04-26
Add support for Leica DICOMs

jcupitt:
This PR adds support for the sample Leica DICOM images.

- use a table to map row-major tile positions to DICOM frame numbers
- test DimensionOrganizationType and for "3D" use PerFrameFunctionalGroupsSequence to construct this table
- support uncompressed RGB pixel data
- support RGB JPEG pixel data

### DIFF
--- a/src/openslide-vendor-dicom.c
+++ b/src/openslide-vendor-dicom.c
@@ -41,6 +41,11 @@
 #define debug(...)
 #endif
 
+enum image_format {
+  FORMAT_JPEG,
+  FORMAT_RGB,
+};
+
 struct dicom_file {
   char *filename;
 
@@ -48,6 +53,7 @@ struct dicom_file {
   DcmFilehandle *filehandle;
   DcmDataSet *metadata;
   DcmBOT *bot;
+  enum image_format format;
 };
 
 struct dicom_level {
@@ -133,6 +139,7 @@ static const char BitsAllocated[] = "BitsAllocated";
 static const char BitsStored[] = "BitsStored";
 static const char HighBit[] = "HighBit";
 static const char PixelRepresentation[] = "PixelRepresentation";
+static const char LossyImageCompression[] = "LossyImageCompression";
 static const char LossyImageCompressionMethod[] = "LossyImageCompressionMethod";
 static const char ObjectiveLensPower[] = "ObjectiveLensPower";
 static const char DimensionOrganizationType[] = "DimensionOrganizationType";
@@ -367,6 +374,15 @@ static void destroy(openslide_t *osr) {
   g_free(osr->levels);
 }
 
+static void rgb_to_cairo(const uint8_t *rgb, uint32_t *dest,
+                         int64_t width, int64_t height) {
+  int64_t n_pixels = width * height;
+  for (int64_t i = 0; i < n_pixels; i++) {
+    dest[i] = 0xff000000 | rgb[0] << 16 | rgb[1] << 8 | rgb[2];
+    rgb += 3;
+  }
+}
+
 static bool decode_frame(struct dicom_file *file, int64_t frame_number,
                          uint32_t *dest, int64_t w, int64_t h,
                          GError **err) {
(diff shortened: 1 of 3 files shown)


bgilbert:
Is an appropriately-licensed demo slide available?

bgilbert:
I've spun out the pin bump into https://github.com/openslide/openslide/pull/442 to deal with the openslide-winbuild CI ratchet.

jcupitt:
> Is an appropriately-licensed demo slide available?

Yes, from the same set as last time, I'll pick one for you.

src/openslide-vendor-dicom.c
```diff
@@ -536,6 +568,19 @@ static bool is_type(char **type, const struct allowed_types *types) {
...
 
+static void rgb_to_cairo(const char *rgb,
+			 uint32_t *dest,
+			 int64_t width,
+			 int64_t height)
+{
+  int64_t n_pixels = width * height;
+
+  for (int64_t i = 0; i < n_pixels; i++) {
+    dest[i] = 0xff000000 | rgb[0] << 16 | rgb[2] << 8 | rgb[2];
```
bgilbert, line 579: Was this meant to be `rgb[1] << 8`?

jcupitt: Ooops! Nice find.

jcupitt:
This looks like the best test image:

ftp://medical.nema.org/MEDICAL/Dicom/DataSets/WG26/Leica_GT450/272320e7-5abb-ccac-8f8b-9c2008f53845_114626

A DICOM written by the Leica gt450 scanner, released under the usual nema terms.

jcupitt:
Here's a shot of that DICOM being viewed via openslide using this PR:

image

src/openslide-vendor-dicom.c
bgilbert: Thanks for the fix.  Please don't add merge commits to PR branches, though.  Something like `git fetch jcupitt add-dicom-sparse-tile-support && git reset --hard FETCH_HEAD` before adding the commit (or perhaps `git pull --rebase --interactive` if you want to preserve some existing commits) should pull new changes into your branch.

bgilbert: (And in fact, it looks like the merge commit reintroduced a [line](https://github.com/openslide/openslide/compare/1f5509582ca873d3d0be905a9e9794b40eb30f56..3f0d44005104ac8d7d5220013700de38638e2a19) that had been removed.)

bgilbert:
The CI failure is probably because the 3DHISTECH DICOM slide is included in frozen test data now.  I'll regenerate it when we're a bit further along.

bgilbert:
> A DICOM written by the Leica gt450 scanner, released under the usual nema terms.

Hmm, that slide is almost 2 GiB, which is large for a test case.  Also, do you have a citation for the license?  I've been relying on [this license statement](https://github.com/DICOMWG26AnnotationAHG/hackathon/tree/main#data-use), which only covers the hackathons, and [this list of datasets](https://github.com/DICOMWG26AnnotationAHG/hackathon/blob/main/data.md).  That slide is sitting in a top-level `Leica_GT450` directory and I'm not 100% sure what license applies.

I thought this slide looked a bit more promising:

ftp://medical.nema.org/MEDICAL/Dicom/DataSets/WG26/WG26Connectathon2022_PV/Leica/Case-D/9cbe447c-1821-610a-8da9-325474ae6d9b_073017/

It's only 100 MB and appears to also have been scanned on a GT450, but libdicom won't read it:

```
libdicom Invalid parameter: Data Element capacity check failed - Value of Data Element 'B63BA242' exceeds maximum length of Value Representation
```

Do you have any other leads on a suitable test slide?

jcupitt:
I think that Leica DICOM is out of spec. It's this element:

```
(0008,002a) DT [20221007073017+429496728900]            #  28, 1 AcquisitionDateTime
```

(from `dcmdump`, the standard DICOM dumper).

[The spec says](https://dicom.nema.org/dicom/2013/output/chtml/part05/sect_6.2.html) (scroll down to DT) that these values should be:

```
YYYYMMDDHHMMSS.FFFFFF&ZZXX
20221007073017+429496728900
```

Where `.FFFFFF` is optional and `&` can also be `+`. It looks like they've put far too much into the `FFF` part and overflowed.

I suppose this should probably be a warning rather than a hard error, I'll change it.

jcupitt:
... git main libdicom can now read this Leica DICOM. Phew!

jcupitt:
openslide displaying that sample DICOM:

image

bgilbert:
Test case added!

Issue #139, opened 2014-08-13
Add API for reading slide ICC profiles


bgilbert:
[Discussion](http://lists.andrew.cmu.edu/pipermail/openslide-users/2014-August/000904.html).

dvdmena:
Connection with https://github.com/openslide/openslide/issues/237

Any news about?

lukenovak:
@bgilbert would you have any recommendations for the general form of this API? I know there were concerns about keeping the profile in memory with `*openslide_get_profile(openslide_t *osr, size_t *length);`, but was curious if that was still a potential issue.

bgilbert:
Yes, I'm told that some formats have multi-megabyte ICC profiles.  The application should provide the buffer, as it does with `openslide_read_region` and `openslide_read_associated_image`, but it'll need a way to get the buffer size.  I see a few possible approaches:

1. `uint64_t openslide_read_icc_profile(openslide_t *osr, void *dest, uint64_t len)`, where OpenSlide writes up to `len` bytes and returns the number of bytes it wanted to write.  That API is compact in some sense, but confusing, and requires the application to call the function twice.
2. `void openslide_read_icc_profile(openslide_t *osr, void *dest, uint64_t len)` and `uint64_t openslide_get_icc_profile_size(openslide_t *osr)`.  This is consistent with e.g. `openslide_get_level_dimensions`.
3. `void openslide_read_icc_profile(openslide_t *osr, void *dest, uint64_t len)`, where the profile size is stored in a property.  The application would have to convert the size property to an integer.

(`size_t` is traditional, but OpenSlide has managed to keep CPU-dependent field widths out of its public API, and I'm hesitant to start using them now.)

I think the right approach is probably 2.  We often _also_ put API-accessible metadata into properties, so that e.g. `openslide-show-properties` gives a complete picture, though notably we don't do that with associated images.

bgilbert:
...and similarly for associated images (`openslide_get_associated_image_icc_profile_size` / `openslide_read_associated_image_icc_profile`), if we want to support profiles there.

We could perhaps drop the `len` argument, since the application knows how to size the buffer.  Notably the other `read` functions don't have one.

lukenovak:
Good to know, and I think 2 or 3 are viable solutions. If 2 seems best to you, then that's probably the way forward. I don't think we'll need the `len` argument, or if we do we won't need to use it for tifflike slides.

[...]

PR #445, opened 2023-05-09
Add a tool to query and write associated images.

alex-virodov:
This is part of tooling that I need to implement request https://github.com/openslide/openslide/pull/437#discussion_r1174073883.

Tested:
```
~/openslide/builddir$ tools/openslide-write-associated-images /home/me/Downloads/TCGA-86-A456-01A-01-TSA.996EFBE2-10E2-454B-B4E7-981FA24404B0.svs 
Opening slide: '/home/me/Downloads/TCGA-86-A456-01A-01-TSA.996EFBE2-10E2-454B-B4E7-981FA24404B0.svs'
Listing associated images:
associated image: 'thumbnail'
Done listing.

~/openslide/builddir$ tools/openslide-write-associated-images /home/me/Downloads/TCGA-86-A456-01A-01-TSA.996EFBE2-10E2-454B-B4E7-981FA24404B0.svs thumbnail /home/me/openslide/thumbnail.png
Opening slide: '/home/me/Downloads/TCGA-86-A456-01A-01-TSA.996EFBE2-10E2-454B-B4E7-981FA24404B0.svs'
Extracting associated image: 'thumbnail'
Dimensions: 1024, 217
Writing output file: '/home/me/openslide/thumbnail.png'
```
----
Note: the `openslide_close()` fails for some reason on SIGSEGV, trying to `g_free(osr);`. Is this a known bug? I don't see the openslide-write-png closing the openslide handle either. If I add `openslide_close(osr)` to the sample, I get the same failure:
```
/home/me/openslide/builddir/tools/openslide-write-png /home/me/Downloads/TCGA-86-A456-01A-01-TSA.996EFBE2-10E2-454B-B4E7-981FA24404B0.svs 0 0 3 1000 1000 /home/me/openslide/output.png

Process finished with exit code 139 (interrupted by signal 11: SIGSEGV)
```

### DIFF
--- a/tools/meson.build
+++ b/tools/meson.build
@@ -8,6 +8,7 @@ tools = [
   ['openslide-quickhash1sum', tools_deps],
   ['openslide-show-properties', tools_deps],
   ['openslide-write-png', [tools_deps, png_dep]],
+  ['openslide-write-associated-images', [tools_deps, png_dep]],
 ]
 foreach t : tools
   executable(
--- a/tools/openslide-write-associated-images.1.in
+++ b/tools/openslide-write-associated-images.1.in
@@ -1,0 +1,67 @@
+.\"
+.\" OpenSlide, a library for reading whole slide image files
+.\"
+.\" Copyright (c) 2007-2012 Carnegie Mellon University
+.\" Copyright (c) 2023 Alexandr Virodov
+.\" All rights reserved.
+.\"
+.\" OpenSlide is free software: you can redistribute it and/or modify
+.\" it under the terms of the GNU Lesser General Public License as
+.\" published by the Free Software Foundation, version 2.1.
+.\"
+.\" OpenSlide is distributed in the hope that it will be useful,
+.\" but WITHOUT ANY WARRANTY; without even the implied warranty of
+.\" MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+.\" GNU Lesser General Public License for more details.
+.\"
+.\" You should have received a copy of the GNU Lesser General Public
+.\" License along with OpenSlide. If not, see
+.\" <http://www.gnu.org/licenses/>.
+.\"
+
+
+.\" See man-pages(7) for formatting conventions.
+
+
+.TH OPENSLIDE-WRITE-ASSOCIATED-IMAGES 1 2023-05-09 "OpenSlide @SUFFIXED_VERSION@" "User Commands"
+
+.mso www.tmac
+
+.SH NAME
+openslide-write-associated-images \- Query or write associated images  of a virtual slide.
+
+.SH SYNOPSIS
+.BR "openslide-write-associated-images " [ --help "] [" --version ]
+.I slide-file - to list associated images.
+.I slide-file associated_image output.png - to write an associated image.
+
+.SH DESCRIPTION
+Query a list of associated images, or write a single associated image to a PNG file.
+
+.SH OPTIONS
+.TP
+.B --help
+Display usage summary.
+
+.TP
+.B --version
+Display version and copyright information.
+
+.SH EXIT STATUS
+.B openslide-write-associated-images
+returns 0 on success, 1 if an error occurred, or 2 if the arguments are
+invalid.
+
+.SH COPYRIGHT
+Copyright \(co 2007-2023 Carnegie Mellon University and others
+
+OpenSlide is free software: you can redistribute it and/or modify it under
+the terms of the
+.URL http://gnu.org/licenses/lgpl-2.1.html "GNU Lesser General Public License, version 2.1" .
+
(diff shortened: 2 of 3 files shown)


bgilbert:
Thanks for the PR!

It feels slightly awkward to have two tools that do almost the same thing, and that share a lot of duplicated code.  From the name of `openslide-write-png`, it'd make sense for that tool to handle associated images as well.  I'd propose adding an `--associated` option to `openslide-write-png`, which would cause it to take an argument list of `slide associated-image output.png`.  Unfortunately that would require some refactoring of `openslide-common-cmdline.c`, since there's currently no way for tools to add additional options to the `GOptionContext`.

As for listing the associated images... I'm torn.  Adding an option to `openslide-show-properties` makes some sense, except that the list of associated images isn't a property.  Adding the option to `openslide-write-png` makes sense to keep the image-related utilities together, except that the list of associated images isn't a PNG.

Interested in your thoughts here.

The crash you're describing is because an explicit `openslide_close(osr)` would cause a double free.  The `openslide_t` is already being closed automatically by the `g_autoptr(openslide_t)`.

alex-virodov:
> Interested in your thoughts here.

My thought is that the proposed structure is the optimal one of course :). More seriously:

The proposed code follows the unix philosophy of "Write programs that do one thing and do it well". This tool handles all aspects of associated images, and does not handle other aspects handled by other tools. So I respectfully disagree with "two tools that do almost the same thing" - they do two very different things imho, just happen to write .pngs in the process.

As to code duplication, the only real code duplication is the 'fail' function and *some* png write logic (as both tools write pngs, but very differently). If there's a place to refactor that code to, I can do the refactor, but I think the write_png function should stay as-is, and only the fail() function should be refactored.

Those are my thoughts. I am happy to continue to discuss, or follow any specific direction you would have.

bgilbert:
OpenSlide is primarily a library.  It's good to provide a command-line tool for exercising it, but I'd rather not keep adding separate small programs for each piece of the API, that downstreams will then need to package.  (One practical problem is that the Windows binaries are now statically linked and cost about 1.5 megabytes each in the zipfile.)

For example, there's recently been some discussion of implementing #139.  Should we add a separate dedicated program just to extract an ICC profile?  It doesn't really fit into any of the existing ones.

So actually, the best approach might be to ship a single program e.g. `openslide-tool` with subcommands like `openslide-tool read-region` and `openslide-tool properties`.  This PR would then add e.g. `openslide-tool associated list` and `openslide-tool associated read`.  The command could also check `argv[0]` and act as the legacy tools if invoked by those names.

If you'd like to work on that, feel free, but I don't expect you to embark on a major refactoring.  We can leave this PR open as a placeholder, and I may try out that approach and see how it feels.  If we're going to make that change, 4.0.0 would be a good time.

In any event, this shouldn't be necessary for #437, so let's keep it independent of that PR.  I usually use OpenSlide Python to test the API.

alex-virodov:
> One practical problem is that .. binaries .. cost about 1.5 megabytes each

I understand completely, and agree with the rest of the points.

> In any event, this shouldn't be necessary for https://github.com/openslide/openslide/pull/437

I will remove it from that branch.

>  I usually use OpenSlide Python to test the API.

FYI one downside of this approach is that you cannot use a C debugger, as you won't have a C entry point. At least not conveniently, afaik (*). Not always needed, but sometimes useful. Maybe an alternative would be to have this program as a testing tool so it won't be shipped.

(*) If you know of a convenient way to debug C code from python entrypoint with CLion, I would be curious to just know that it is possible. My experience is with Java/JNI/C++, and it is not a good one.

bgilbert:
> FYI one downside of this approach is that you cannot use a C debugger, as you won't have a C entry point.

Hmm, works for me:

```
$ gdb python

[elided prompts about debuginfod and loading debuginfo]

(gdb) break openslide_read_associated_image
Function "openslide_read_associated_image" not defined.
Make breakpoint pending on future shared library load? (y or [n]) y
Breakpoint 1 (openslide_read_associated_image) pending.
(gdb) run
Starting program: /usr/bin/python 

[elided debuginfo messages]

Python 3.11.3 (main, Apr  5 2023, 00:00:00) [GCC 13.0.1 20230401 (Red Hat 13.0.1-0)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import openslide

[elided debuginfo messages]

>>> s=openslide.OpenSlide('CMU-1.svs')
>>> s.associated_images['label']

Breakpoint 1, openslide_read_associated_image (osr=0x55555593cd30, 
    name=0x7fffe9fac2c0 "label", dest=0x7fffe879a010) at ../src/openslide.c:572
572	  if (openslide_get_error(osr)) {
(gdb) 
```

bgilbert:
I've implemented the subcommand approach in #465.  It's still a draft, but feels like a good direction to go in.

I'll go ahead and close this PR for now.  Thanks for sending it, and for pointing out the need for this tooling!

PR #454, opened 2023-06-02
Revise DICOM loader for libdicom 0.4

jcupitt:
libdicom 0.4 has a simplified API that lets us remove quite a bit of code.

libdicom 0.4 is quite a bit quicker and needs less memory. For example, here's openslide plus libdicom 0.2 on a large Leica DICOM:

```
$ /usr/bin/time -f %M:%e openslide-show-properties 272320e7-5abb-ccac-8f8b-9c2008f53845_114626/1.3.6.1.4.1.36533.1214640142408762111756320324051133413.dcm
openslide.level-count: '4'
openslide.level[0].downsample: '1'
openslide.level[0].height: '84441'
openslide.level[0].tile-height: '256'
openslide.level[0].tile-width: '256'
openslide.level[0].width: '108071'
openslide.level[1].downsample: '4.0000792060552719'
openslide.level[1].height: '21110'
openslide.level[1].tile-height: '256'
openslide.level[1].tile-width: '256'
openslide.level[1].width: '27017'
openslide.level[2].downsample: '16.0013709686787'
openslide.level[2].height: '5277'
openslide.level[2].tile-height: '256'
openslide.level[2].tile-width: '256'
openslide.level[2].width: '6754'
openslide.level[3].downsample: '64.021029009122955'
openslide.level[3].height: '1319'
openslide.level[3].tile-height: '256'
openslide.level[3].tile-width: '256'
openslide.level[3].width: '1688'
openslide.mpp-x: '0.26297895237999996'
openslide.mpp-y: '0.26297895237999996'
openslide.vendor: 'dicom'
691592:4.15
```

700mb of memory and 4.2s.

With this PR:

```
$ /usr/bin/time -f %M:%e openslide-show-properties 272320e7-5abb-ccac-8f8b-9c2008f53845_114626/1.3.6.1.4.1.36533.21716722913121316773164147287613518920414871.dcm
openslide.level-count: '4'
openslide.level[0].downsample: '1'
openslide.level[0].height: '84441'
openslide.level[0].tile-height: '256'
openslide.level[0].tile-width: '256'
openslide.level[0].width: '108071'
openslide.level[1].downsample: '4.0000792060552719'
openslide.level[1].height: '21110'
openslide.level[1].tile-height: '256'
openslide.level[1].tile-width: '256'
openslide.level[1].width: '27017'
openslide.level[2].downsample: '16.0013709686787'
openslide.level[2].height: '5277'
openslide.level[2].tile-height: '256'
openslide.level[2].tile-width: '256'
openslide.level[2].width: '6754'
openslide.level[3].downsample: '64.021029009122955'
openslide.level[3].height: '1319'
openslide.level[3].tile-height: '256'
openslide.level[3].tile-width: '256'
openslide.level[3].width: '1688'
openslide.mpp-x: '0.26297895237999996'
openslide.mpp-y: '0.26297895237999996'
openslide.vendor: 'dicom'
89560:0.04
```

90mb of memory and 0.04s.

### DIFF
--- a/src/openslide-decode-dicom.c
+++ b/src/openslide-decode-dicom.c
@@ -24,6 +24,12 @@
 #include "openslide-private.h"
 #include "openslide-decode-dicom.h"
 
+// implements DcmIO
+struct _dicom_io {
+  DcmIOMethods *methods;
+  struct _openslide_file *file;
+};
+
 void _openslide_dicom_propagate_error(GError **err, DcmError *dcm_error) {
   g_set_error(err, OPENSLIDE_ERROR, OPENSLIDE_ERROR_FAILED,
               "libdicom %s: %s - %s",
@@ -40,45 +46,49 @@ static void propagate_gerror(DcmError **dcm_error, GError *err) {
   g_error_free(err);
 }
 
-static void *vfs_open(DcmError **dcm_error, void *client) {
+static DcmIO *vfs_open(DcmError **dcm_error, void *client) {
   const char *filename = (const char *) client;
+  struct _dicom_io *dio = g_new(struct _dicom_io, 1);
 
   GError *err = NULL;
-  struct _openslide_file *file = _openslide_fopen(filename, &err);
-  if (!file) {
+  dio->file = _openslide_fopen(filename, &err);
+  if (!dio->file) {
+    g_free(dio);
     propagate_gerror(dcm_error, err);
     return NULL;
   }
 
-  return file;
+  return (DcmIO *) dio;
 }
 
-static int vfs_close(DcmError **dcm_error G_GNUC_UNUSED, void *data) {
-  struct _openslide_file *file = (struct _openslide_file *) data;
-  _openslide_fclose(file);
-  return 0;
+static void vfs_close(DcmIO *io) {
+  struct _dicom_io *dio = (struct _dicom_io *) io;
+
+  _openslide_fclose(dio->file);
+  g_free(dio);
 }
 
-static int64_t vfs_read(DcmError **dcm_error G_GNUC_UNUSED, void *data,
+static int64_t vfs_read(DcmError **dcm_error G_GNUC_UNUSED, DcmIO *io,
                         char *buffer, int64_t length) {
-  struct _openslide_file *file = (struct _openslide_file *) data;
+  struct _dicom_io *dio = (struct _dicom_io *) io;
+
   // openslide VFS has no error return for read()
-  return _openslide_fread(file, buffer, length);
+  return _openslide_fread(dio->file, buffer, length);
 }
 
-static int64_t vfs_seek(DcmError **dcm_error, void *data,
+static int64_t vfs_seek(DcmError **dcm_error, DcmIO *io,
                         int64_t offset, int whence) {
-  struct _openslide_file *file = (struct _openslide_file *) data;
+  struct _dicom_io *dio = (struct _dicom_io *) io;
 
   GError *err = NULL;
-  if (!_openslide_fseek(file, offset, whence, &err)) {
+  if (!_openslide_fseek(dio->file, offset, whence, &err)) {
     propagate_gerror(dcm_error, err);
     return -1;
   }
 
   // libdicom uses lseek(2) semantics, so it must always return the new file
   // pointer
-  off_t new_position = _openslide_ftell(file, &err);
+  off_t new_position = _openslide_ftell(dio->file, &err);
   if (new_position < 0) {
     propagate_gerror(dcm_error, err);
   }
@@ -86,7 +96,7 @@ static int64_t vfs_seek(DcmError **dcm_error, void *data,
   return new_position;
 }
 
-static const DcmIO io_funcs = {
+static const DcmIOMethods dicom_open_methods = {
   .open = vfs_open,
   .close = vfs_close,
   .read = vfs_read,
@@ -94,12 +104,20 @@ static const DcmIO io_funcs = {
 };
 
 DcmFilehandle *_openslide_dicom_open(const char *filename, GError **err) {
+
   DcmError *dcm_error = NULL;
-  DcmFilehandle *result =
-    dcm_filehandle_create(&dcm_error, &io_funcs, (void *) filename);
-  if (!result) {
+  DcmIO *io = dcm_io_create(&dcm_error, &dicom_open_methods, (void *) filename);
+  if (!io) {
     _openslide_dicom_propagate_error(err, dcm_error);
     return NULL;
   }
-  return result;
+
+  DcmFilehandle *filehandle = dcm_filehandle_create(&dcm_error, io);
+  if (!filehandle) {
+    _openslide_dicom_propagate_error(err, dcm_error);
+    dcm_io_close(io);
+    return NULL;
+  }
+
+  return filehandle;
 }
--- a/src/openslide-vendor-dicom.c
+++ b/src/openslide-vendor-dicom.c
@@ -52,7 +52,6 @@ struct dicom_file {
   GMutex lock;
   DcmFilehandle *filehandle;
   DcmDataSet *metadata;
-  DcmBOT *bot;
   enum image_format format;
 };
 
@@ -60,14 +59,9 @@ struct dicom_level {
   struct _openslide_level base;
   struct _openslide_grid *grid;
 
-  int64_t tiles_across;
-  int64_t tiles_down;
-
   double pixel_spacing_x;
   double pixel_spacing_y;
   double objective_lens_power;
-
-  int64_t *frame_position;
 
   struct dicom_file *file;
 };
@@ -142,17 +136,12 @@ static const char PixelRepresentation[] = "PixelRepresentation";
 static const char LossyImageCompression[] = "LossyImageCompression";
 static const char LossyImageCompressionMethod[] = "LossyImageCompressionMethod";
 static const char ObjectiveLensPower[] = "ObjectiveLensPower";
-static const char DimensionOrganizationType[] = "DimensionOrganizationType";
-static const char PerFrameFunctionalGroupsSequence[] = "PerFrameFunctionalGroupsSequence";
-static const char FrameContentSequence[] = "FrameContentSequence";
-static const char DimensionIndexValues[] = "DimensionIndexValues";
 
 static void print_file(struct dicom_file *f G_GNUC_UNUSED) {
   debug("file:" );
   debug("  filename = %s", f->filename);
   debug("  filehandle = %p", f->filehandle);
   debug("  metadata = %p", f->metadata);
-  debug("  bot = %p", f->bot);
 }
 
 static void print_level(struct dicom_level *l) {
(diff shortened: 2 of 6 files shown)


bgilbert:
If we're not maintaining our own tile position mappings anymore, the call to `_openslide_grid_draw_tile_info()` is no longer useful and could be dropped.

jcupitt:
The test failure seems to be caused because `dicom.jpg`, the sample DICOM in `openslide-vendor-synthetic.c`, is missing the `TotalMatrixPixelColumns` ta:

https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.8.12.14.html#sect_C.8.12.14.1.1

And libdicom currently assumes that tag is present. 

Do you remember where that sample DICOM came from? Perhaps I could test it in one of the online DICOM standard checkers.

bgilbert:
It was generated with:

    convert -quality 90 misc/synthetic.png synthetic.jpg
    img2dcm -nsc synthetic.jpg synthetic.dcm

[Here's a copy](https://github.com/openslide/openslide/files/11643155/synthetic.dcm.gz).

jcupitt:
I updated for the new `read_frame_position` API call and made 0x00480006 optional in libdicom.

src/openslide-vendor-dicom.c
```diff
@@ -490,7 +469,7 @@ static bool read_tile(openslide_t *osr,
                                         l->base.tile_w * 4);
   cairo_set_source_surface(cr, surface, 0, 0);
   cairo_paint(cr);
-  _openslide_grid_draw_tile_info(cr, "%"PRId64, frame_number);
+  _openslide_grid_draw_tile_info(cr, "%"PRId64",%"PRId64, tile_col, tile_row);
```
bgilbert, line 472: The grid already draws this info automatically, so let's drop this.

src/openslide-decode-dicom.c
```diff
@@ -33,73 +33,89 @@ void _openslide_dicom_propagate_error(GError **err, DcmError *dcm_error) {
   dcm_error_clear(&dcm_error);
 }
 
+// implements DcmIO
+struct _dicom_io {
```
bgilbert, line 37: Nit: this is awkwardly placed between the two error-handling functions.  Let's move it above `_openslide_dicom_propagate_error`?

src/openslide-vendor-dicom.c
```diff
@@ -681,84 +660,6 @@ static bool add_level(openslide_t *osr,
...
-      if (l->frame_position[frame_index] != -1) {
-        g_set_error(err, OPENSLIDE_ERROR, OPENSLIDE_ERROR_FAILED,
-                    "Duplicate frame (%"PRId64", %"PRId64")", col, row);
-        return false;
-      }
-      l->frame_position[frame_index] = i + 1;
-    }
-  } else {
-    g_set_error(err, OPENSLIDE_ERROR, OPENSLIDE_ERROR_FAILED,
-                "Unsupported DimensionOrganisationType");
```
bgilbert, line 758: Is libdicom now doing this check internally?

src/openslide-vendor-synthetic.c
```diff
@@ -124,14 +124,8 @@ static bool decode_dicom(const void *data, uint32_t len,
...
   }
-  g_autoptr(DcmBOT) bot = dcm_filehandle_read_bot(&dcm_error, fh, metadata);
-  if (!bot) {
-    _openslide_dicom_propagate_error(err, dcm_error);
-    g_prefix_error(err, "Reading BOT: ");
-    return false;
-  }
   g_autoptr(DcmFrame) frame =
-    dcm_filehandle_read_frame(&dcm_error, fh, metadata, bot, 1);
+    dcm_filehandle_read_frame(&dcm_error, fh, 1);
```
bgilbert, line 128: We could also drop the `dcm_filehandle_read_metadata` call, right?

src/openslide-vendor-dicom.c
```diff
@@ -442,24 +431,14 @@ static bool read_tile(openslide_t *osr,
   debug("read_tile level:");
   print_level(l);
 
-  int64_t frame_index = tile_col + l->tiles_across * tile_row;
```
bgilbert, line 445: This is the last user of `l->tiles_across`, and `l->tiles_down` already didn't have any, so we could drop those fields.

src/openslide-decode-dicom.c
```diff
@@ -33,73 +33,89 @@ void _openslide_dicom_propagate_error(GError **err, DcmError *dcm_error) {
...
 
-static const DcmIO io_funcs = {
-  .open = vfs_open,
-  .close = vfs_close,
-  .read = vfs_read,
-  .seek = vfs_seek,
-};
-
 DcmFilehandle *_openslide_dicom_open(const char *filename, GError **err) {
+  static const DcmIOMethods methods = {
```
bgilbert, line 99: Style nit: we usually place these outside of functions, since they do leak outside the function's scope.

src/openslide-decode-dicom.c
```diff
@@ -33,73 +33,89 @@ void _openslide_dicom_propagate_error(GError **err, DcmError *dcm_error) {
...
   const char *filename = (const char *) client;
+  struct _dicom_io *dio = g_new(struct _dicom_io, 1);
 
   GError *err = NULL;
-  struct _openslide_file *file = _openslide_fopen(filename, &err);
-  if (!file) {
+  dio->file = _openslide_fopen(filename, &err);
+  if (!dio->file) {
     propagate_gerror(dcm_error, err);
     return NULL;
```
bgilbert, line 57: Leaks `dio`.  For consistency, let's put a `g_autofree` on `dio` and use `g_steal_pointer()` on return.

bgilbert (commented):
This looks good.  Nice cleanup!  I've added a few minor comments.

It looks like associated images in the 3DHISTECH test slide are failing to read with:

```
libdicom Parse error: Reading Frame position failed - No Frame at position (0, 0)
```

src/openslide-decode-dicom.c
bgilbert: Actually, `g_autofree` isn't right either, since there's a field that would be leaked if we added any additional fallible initialization.  It's probably not worth a dedicated cleanup function for one call site, so I guess we just do the cleanup inline for now.

src/openslide-vendor-dicom.c
jcupitt: Oh nice, done.

src/openslide-vendor-dicom.c
jcupitt: OK, I put this check back.

src/openslide-vendor-synthetic.c
jcupitt: It's needed to scan the first part of the header (it side-effects the filehandle), which is a little ugly. 

This should probably be fixed in a later libdicom.

src/openslide-vendor-dicom.c
jcupitt: Oh yeah, nice.

src/openslide-decode-dicom.c
jcupitt: OK, moved.

src/openslide-decode-dicom.c
jcupitt: Ooop, good catch. I added a `g_free()` on the error path.

jcupitt:
Ah perhaps not.

I've downloaded all the tests and also added your PR for better error detection. It fails if I run:

```
john@banana ~/GIT/openslide/build/test (update-to-libdicom-0.4) $ ./driver run
aperio: OK
aperio-33003: OK
aperio-33005: OK
....
debug-help: OK
debug-tiles: OK
dicom-tiled-full-jpeg: extended test failed: Reading associated image "label" failed: libdicom Parse error: Reading Frame position failed - No Frame at position (0, 0)
dicom-tiled-sparse-jpeg: OK
empty-file: OK
...
```

But seems to work if I run it directly:

```
john@banana ~/GIT/openslide/build/test (update-to-libdicom-0.4) $ ./extended _slidedata/unpacked/dicom-tiled-sparse-jpeg/1.3.6.1.4.1.36533.21773233891171386611617621819013191107166.dcm
john@banana ~/GIT/openslide/build/test (update-to-libdicom-0.4) $ echo $?
0
```

And I think I can read out the label image OK if I run in a debugger. Any idea how I can get this failure locally? I'm doing something dumb.

bgilbert:
Looks like you're manually running it against `dicom-tiled-sparse-jpeg`, but the problem is in `dicom-tiled-full-jpeg`:

```console
$ ./extended ~/openslide/_slidedata/unpacked/dicom-tiled-full-jpeg/000009.dcm 
Read failed: -10 -10 200 200: libdicom Parse error: Reading Frame position failed - No Frame at position (0, 0)
```

src/openslide-vendor-synthetic.c
bgilbert: Fair enough.  Can we add a comment to that effect above the `read_metadata` call so the purpose is clear?

bgilbert (commented):
Fixups look good!

src/openslide-decode-dicom.c
```diff
@@ -24,6 +24,12 @@
 #include "openslide-private.h"
 #include "openslide-decode-dicom.h"
 
+// implements DcmIO
+struct _dicom_io {
+    DcmIOMethods *methods;
```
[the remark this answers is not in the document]
jcupitt: OK, done.

jcupitt:
OK, leak fixed! Phew, thanks for the reviews.

src/openslide-vendor-synthetic.c
bgilbert: Ah, I meant on the call in `openslide-vendor-synthetic.c`.  The one in `openslide-vendor-dicom.c` doesn't need a comment because we're actually using the returned metadata.

src/openslide-vendor-dicom.c
bgilbert: Looks like it's gone again, I guess because of [this code](https://github.com/jcupitt/libdicom/commit/4828495d3c5c325fb258b9df915f7948e8d4fbde#diff-e4206ff375a29a4fc3867522531849694c83dddf198db2b3d602b3b0261d6decR709-R713)?  I agree that we should be trusting libdicom to know what it can and cannot read, rather than imposing our own judgement in OpenSlide.  But my reading of the libdicom code is that if the `DimensionOrganizationType` is something unexpected like `3D_TEMPORAL`, we'll try to compute the tile index as though it were `TILED_FULL`, and return random tiles.  Could we have one or the other library fail if the organization type is unrecognized?

bgilbert (commented):
The CI failure on macOS appears to be because clang isn't happy with the redefinition of the `DcmIOMethods` typedef in `dicom.h`.

src/openslide-vendor-synthetic.c
jcupitt: Ah! I was a little puzzled, OK, done.

bgilbert (approved):
LGTM!  This PR is still in draft state; is it ready to merge?

jcupitt:
I was waiting for libdicom to merge 0.4 to main, but perhaps we don't have time for that. I'll merge 0.4 to main in my fork and then update this PR to use that main.

jcupitt:
OK, I merged libdicom 0.4 to main on jcupitt/libdicom and updated `libdicom.wrap` to point to that.

jcupitt:
Thanks!

PR #456, opened 2023-06-08
Get dicom objective-power from optical-path

jcupitt:
The objective-power metadata item is defined in the optical path sequence:

https://dicom.innolitics.com/ciods/vl-microscopic-image/optical-path/00480105/00480112

This PR fetches objective power from that sequence, not from the main metadata dataset.

With this PR, openslide-show-properties gets the OP correctly from the sample DICOMs I have:

```
$ openslide-show-properties 1.3.6.1.4.1.36533.116129230228107214763613716719238114924751.dcm | grep -i obj
openslide.objective-power: '40'
```

### DIFF
--- a/src/openslide-vendor-dicom.c
+++ b/src/openslide-vendor-dicom.c
@@ -135,6 +135,7 @@ static const char HighBit[] = "HighBit";
 static const char PixelRepresentation[] = "PixelRepresentation";
 static const char LossyImageCompression[] = "LossyImageCompression";
 static const char LossyImageCompressionMethod[] = "LossyImageCompressionMethod";
+static const char OpticalPathSequence[] = "OpticalPathSequence";
 static const char ObjectiveLensPower[] = "ObjectiveLensPower";
 
 static void print_file(struct dicom_file *f G_GNUC_UNUSED) {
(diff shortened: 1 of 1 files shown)


src/openslide-vendor-dicom.c
```diff
@@ -826,6 +833,8 @@ static void add_properties(openslide_t *osr, struct dicom_level *level0) {
                         g_strdup(OPENSLIDE_PROPERTY_NAME_OBJECTIVE_POWER),
                         _openslide_format_double(level0->objective_lens_power));
   }
+
+
```
bgilbert, line 837: Stray added lines?

src/openslide-vendor-dicom.c
```diff
@@ -673,7 +674,13 @@ static bool add_level(openslide_t *osr,
     get_tag_decimal_str(pixel_measures, PixelSpacing, 1, &l->pixel_spacing_y);
   }
   // objective power
-  get_tag_decimal_str(f->metadata, ObjectiveLensPower, 0, &l->objective_lens_power);
+  DcmDataSet *optical_path;
+  if (get_tag_seq_item(f->metadata,
+                       OpticalPathSequence,
+                       0,
+                       &optical_path)) {
```
bgilbert, line 681: Nit: these will all fit on one line so we don't need to stack them.

bgilbert (commented):
LGTM, thanks!  Just a couple typographic nits.  While you're at it, maybe rebase onto `main` so CI runs against libdicom 0.4?

bgilbert (approved):

PR #457, opened 2023-06-08
dicom: support quickhash-1

jcupitt:
The purpose of the quickhash is to `uniquely identify a particular virtual slide, but [not] to detect file corruption or modification`.  Most slide formats don't include unique identifiers, so we usually have to construct one from metadata that collectively is unique to a slide.

In the case of DICOM, there are already unique identifiers for entities at various conceptual levels.  Those identifiers refer to the conceptual entity, not its exact representation.  For example, the SOP Instance UID doesn't change if the pixel data or metadata are altered in certain ways that don't change the instance's notional semantics.

Rather than second-guessing DICOM's approach to entity naming, adopt it.  That requires us to pick a UID to hash.  The Acquisition UID is too general (two recompressions of the same scan share the same UID) and its presence is optional; the Pyramid UID is optional; and the SOP Instance UID refers to a specific level, portion of a level, or associated image.  We could hash all the SOP Instance UIDs, but then we'd be creating our own definition of a "slide" out of its constituent parts.

The Series Instance UID, however, seems like a good fit, especially since we currently support only one series per slide.  Calculate the quickhash by hashing that UID.  In the future, if we support Z-stacks distributed across multiple series (with a common Acquisition ID), we can hash a sorted list of all Series Instance UIDs.

This will cause quickhash collisions if a single series (or a reused series UID) contains multiple independent slides _and_ we aren't presented with the entire series at once.  We could try to avoid that by hashing additional attributes such as timestamps, but again we'd be second-guessing the existing naming semantics.

### DIFF
--- a/src/openslide-vendor-dicom.c
+++ b/src/openslide-vendor-dicom.c
@@ -23,6 +23,8 @@
 
 /*
  * DICOM (.dcm) support
+ *
+ * quickhash comes from the Series Instance UID
  */
 
 /*
@@ -1095,8 +1097,8 @@ static bool dicom_open(openslide_t *osr,
 
   (void) get_icc_profile(level0->file, &osr->icc_profile_size);
 
-  // no quickhash yet; disable
-  _openslide_hash_disable(quickhash1);
+  // compute quickhash
+  _openslide_hash_string(quickhash1, slide_id);
 
   g_assert(osr->data == NULL);
   g_assert(osr->levels == NULL);
--- a/misc/dicom-quickhash.py
+++ b/misc/dicom-quickhash.py
@@ -1,0 +1,10 @@
+#!/usr/bin/python3
+
+import hashlib
+from pydicom import dcmread
+import sys
+
+h = hashlib.sha256()
+h.update(dcmread(sys.argv[1])[(0x20,0xe)].value.encode() + b'\0')
+
+print(h.hexdigest())
--- a/test/cases/dicom-tiled-full-jp2k-rgb/config.yaml
+++ b/test/cases/dicom-tiled-full-jp2k-rgb/config.yaml
@@ -9,4 +9,5 @@ properties:
   openslide.associated.macro.icc-size: "3144"
   openslide.associated.thumbnail.icc-size: "3144"
   openslide.icc-size: "3144"
+  openslide.quickhash-1: 8d3fcfae032e2758633c75fa65a583acf4dd22e0a28872d628947d3937f5c2f8
   openslide.vendor: dicom
--- a/test/cases/dicom-tiled-full-jp2k-ycbcr/config.yaml
+++ b/test/cases/dicom-tiled-full-jp2k-ycbcr/config.yaml
@@ -9,4 +9,5 @@ properties:
   openslide.associated.macro.icc-size: "141992"
   openslide.associated.thumbnail.icc-size: "141992"
   openslide.icc-size: "141992"
+  openslide.quickhash-1: 2431797d260ab5fab926a01f3cee1d35dbb7281f82aed46d97c5aafeff16079c
   openslide.vendor: dicom
--- a/test/cases/dicom-tiled-full-jpeg/config.yaml
+++ b/test/cases/dicom-tiled-full-jpeg/config.yaml
@@ -8,4 +8,5 @@ properties:
   openslide.associated.label.icc-size: "63928"
   openslide.associated.macro.icc-size: "63928"
   openslide.icc-size: "63928"
+  openslide.quickhash-1: 95620a2bd3ca53b8bdc0830fe99ad5f9f150d80e202c5e74aec3c905ebbad0cb
   openslide.vendor: dicom
--- a/test/cases/dicom-tiled-sparse-jpeg/config.yaml
+++ b/test/cases/dicom-tiled-sparse-jpeg/config.yaml
@@ -7,4 +7,5 @@ primary: true
 properties:
   openslide.associated.thumbnail.icc-size: "13113264"
   openslide.icc-size: "13113264"
+  openslide.quickhash-1: 7ca87626aa9842b867f284a4ac0c9474f5de295455c8bb25be2ed837efcc5b10
   openslide.vendor: dicom


src/openslide-vendor-dicom.c
```diff
@@ -711,6 +726,76 @@ static void add_properties(openslide_t *osr, struct dicom_level *level0) {
   }
 }
 
+static bool quickhash_level_pixels(struct _openslide_hash *hash,
+                                   struct dicom_level *level) {
+  double d;
+  if (!get_tag_decimal_str(level->file->metadata, NumberOfFrames, 0, &d)) {
+    return false;
```
bgilbert, line 733: Are there slides where the lowest-resolution level won't have any frames, and we want to skip the quickhash but not fail the open?  It seems like we'd want to fail in that case.

src/openslide-vendor-dicom.c
```diff
@@ -694,7 +708,8 @@ static bool maybe_add_file(openslide_t *osr,
   }
 }
 
-static void add_properties(openslide_t *osr, struct dicom_level *level0) {
+static void add_properties(openslide_t *osr,
+                           struct dicom_level *level0) {
```
bgilbert, line 712: This seems fine as-is.

src/openslide-vendor-dicom.c
```diff
@@ -711,6 +726,76 @@ static void add_properties(openslide_t *osr, struct dicom_level *level0) {
   }
 }
 
+static bool quickhash_level_pixels(struct _openslide_hash *hash,
+                                   struct dicom_level *level) {
+  double d;
+  if (!get_tag_decimal_str(level->file->metadata, NumberOfFrames, 0, &d)) {
```
bgilbert, line 732: Isn't this an Integer String?

src/openslide-vendor-dicom.c
```diff
@@ -711,6 +726,76 @@ static void add_properties(openslide_t *osr, struct dicom_level *level0) {
...
+  StudyDate,
+  StudyTime,
+  PatientID,
+  StudyID,
+  SeriesNumber,
+  ReferringPhysicianName,
+  PatientName,
+  PatientBirthDate,
+  PatientSex,
+  PatientOrientation,
```
bgilbert, line 778: Hmm.  Should we include the content/acquisition date/time in addition to (or in lieu of) the study ones?  If we're including the Patient ID, should we include the Issuer of Patient ID?  Some of the patient and physician metadata feels secondary; should we remove it?  Should we include some UIDs?

@hackermd and @dclunie, can you provide any guidance on a suitable set of attributes here?  We're looking for attributes which, taken together, will uniquely identify a DICOM slide, despite any plausible misbehavior of the operator or software that generated it.

For many slide formats, the available metadata fields can't uniquely identify a slide, so OpenSlide also hashes the compressed pixel data from the lowest-resolution level.  Is that necessary for DICOM?  If there's a combination of metadata attributes which guarantee uniqueness &mdash; in particular, if UID collisions don't happen in the wild &mdash; it'd be great to skip the content hash.

src/openslide-vendor-dicom.c
```diff
@@ -711,6 +726,76 @@ static void add_properties(openslide_t *osr, struct dicom_level *level0) {
...
+  }
+  uint32_t num_frames = d;
+
+  int64_t total_bytes = 0;
+  for (uint32_t i = 0; i < num_frames; i++) {
+    DcmFrame *frame = dcm_filehandle_read_frame(NULL,
+                                                level->file->filehandle,
+                                                i + 1);
+    if (frame == NULL) {
+      return false;
```
bgilbert, line 743: Likewise, should this fail the open?

src/openslide-vendor-dicom.c
```diff
@@ -711,6 +726,76 @@ static void add_properties(openslide_t *osr, struct dicom_level *level0) {
...
+                                   struct dicom_level *level) {
+  double d;
+  if (!get_tag_decimal_str(level->file->metadata, NumberOfFrames, 0, &d)) {
+    return false;
+  }
+  uint32_t num_frames = d;
+
+  int64_t total_bytes = 0;
+  for (uint32_t i = 0; i < num_frames; i++) {
+    DcmFrame *frame = dcm_filehandle_read_frame(NULL,
```
bgilbert, line 739: `g_autoptr` to fix leaks.

src/openslide-vendor-dicom.c
```diff
@@ -711,6 +726,76 @@ static void add_properties(openslide_t *osr, struct dicom_level *level0) {
...
+  PatientSex,
+  PatientOrientation,
+};
+
+static void quickhash_level_metadata(struct _openslide_hash *hash,
+                                     struct dicom_level *level) {
+  hash_int64(hash, level->base.w);
+  hash_int64(hash, level->base.h);
+  hash_int64(hash, level->base.tile_w);
+  hash_int64(hash, level->base.tile_h);
```
bgilbert, line 786: Would it make sense to include the underlying attributes in `quickhash_level_metadata_tags` instead of special-casing the values here?  These struct fields reflect OpenSlide's presentation of the slide, and it's conceivable that in the future they won't always be equal to the underlying attributes.

bgilbert (commented):
Thanks!

src/openslide-vendor-dicom.c
jcupitt: Ooop! Sorry.

src/openslide-vendor-dicom.c
jcupitt: Yes, though a double will hold the full range and it saved writing `get_tag_integer_str()`.

src/openslide-vendor-dicom.c
dclunie: @bgilbert, yes DICOM UIDs would ideally be the most reliable identifiers of unique entities, such as one scan of a slide (recognizing that each scan may result in multiple DICOM image instances (files) including pyramid layers, overview and label).

First, to state the obvious, if you were going to use the pixel data hash of just one layer of the pyramid as a proxy for the set of files that are one scan of the slide, then the DICOM [SOP Instance UID](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.12.html#sect_C.12.1) of the file (instance) containing that same layer would be an equally suitable proxy, since it should be already globally unique. DICOM implementers and toolkits are generally pretty disciplined about creating well-formed and unique UIDs for this purpose and it is always required to be present.

(F)TL;DNR:

More generally, a discussion of which UID nominally identifies a particular entity and which to select for your use case depends on whether you want to identify a single DICOM "file" (layer of a pyramid, a label, or an overview (macro) file), in which case SOP Instance UID (in that file) should be sufficient, or you want to identify the set of DICOM files that constitute a a single scan (acquisition) of a slide, in which Acquisition UID would work, if present (_vide infra_), assuming that the same acquisition was not stored multiple times (e.g., compressed differently) in DICOM format, or you also wanted to distinguish the same pixel data content in different formats (e.g., want to distinguish SVS from DICOM for same slide).

Traditionally, the general purpose [Series Instance UID](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.7.3.html#sect_C.7.3.1) was thought to have been sufficient for identifying the set of files (instances), but that has not always been the case in practice. During somewhat contrived testing scenarios like early [WG26 Connectathons](https://doi.org/10.4103/jpi.jpi_1_18) we saw some implementers reusing the same Series Instance UID for multiple successive scans of the same slide, and you may encounter this in old test objects such as at ftp://medical.nema.org/MEDICAL/Dicom/DataSets/WG26/.

So, because it is theoretically possible for a creator to put more than one scan of a slide in the same series, in [CP 2135](https://dicom.nema.org/medical/dicom/Final/cp2135_ft_addAcquisitionEntityAndUID.pdf) we recently (2022/03/31) added the [Acquisition UID](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.7.10.html#sect_C.7.10.1) and the [Pyramid UID](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.7.11.html#sect_C.7.11.1) to try to address these sorts of problems going forward. Acquisition UID would seem to fit your use case if you are looking for an identifier for the set of files constituting one scan of a slide such as might otherwise be stored in one SVS or TIFF file with multiple IFDs.

Since Acquisition UID (and Pyramid UID) are recent additions (and optional at the attribute or module level) you may not find many vendors producing them. For testing you should be able to find some images with them in [recent](https://github.com/ImagingDataCommons/idc-wsi-conversion/tree/main#Reconversion) SM objects in the [Imaging Data Commons](https://portal.imaging.datacommons.cancer.gov/explore/filters/?Modality=SM) though, since I included support for generating them in [my converter](https://www.dclunie.com/pixelmed/software/javadoc/com/pixelmed/convert/TIFFToDicom.html).

I might try using a unique key that was a combination of Acquisition UID and Series Instance UID, since that would cover the case when Acquisition UID is missing and the case when a derived (set of) images, such as re-downsampled or image processed or newly lossy compressed set, which might have the same Acquisition UID but ideally would have a different Series Instance UID. I might even be tempted to include several other attributes that might disambiguate original vs. derived (sets of) images if someone was foolish enough to put them in the same series, such as Instance Creation Date and Time (also optional), even Image Type, although if you are going to include a hash of the top layer pixel data or use SOP Instance UID of the top layer instance as a proxy anyway, this might not be necessary.

If you do use SOP Instance UID, you don't need to include the higher level entity UIDs, since the scope of uniqueness of a UID is supposed to be global, not hierarchically dependent on parent entities, and child entities are not supposed to have multiple parents in the study/series/instance hierarchy.

Not sure that a pixel data hash alone is sufficient though, since if someone sends the same scan twice but with different UIDs for some reason, or generates synthetic pixel data for testing, your hash of the pixel data at the top level of the pyramid may identify scans that you think are the same but are not from the same files. One way to disambiguate these is to include the SOP Instance UID of the DICOM instance (file) that the pixel data is extracted from in the hash value.

We sometimes see this sort of thing in radiology PACS where people think a hash of the pixel data is unique, but it is not because the pixel data contains a test pattern (like a SMPTE pattern) that may have been generated by different sources and stored with different identifying metadata. So while the hash uniquely identifies the pixel data, it does not uniquely identify the pixeldata+metadata.

src/openslide-vendor-dicom.c
jcupitt: We'd need to give `quickhash_level_pixels()` three exit states: no suitable level (it's too big to hash quickly, I think this could happen), some kind of bad low level failure, and success, which seems awkward.

Putting the first two cases into one and making it disable quickhash seems like a useful simplification to me.

src/openslide-vendor-dicom.c
jcupitt: I don't think this can realistically fail, so making it just disable quickhash seems reasonable. It'd very probably fail on get tile  soon after, so I don't think the error (eg. a corrupt or badly malformed file, perhaps?) would be hidden.

src/openslide-vendor-dicom.c
jcupitt: ooops! Fixed.

src/openslide-vendor-dicom.c
jcupitt: OK, I added a second table for int-valued tags.

src/openslide-vendor-dicom.c
jcupitt: Ah, interesting!

I added the tags you suggest, so the slide hash is now formed from a combination of (if present):

```
  AcquisitionUID,
  InstanceCreationDate,
  InstanceCreationTime,
  PatientBirthDate,
  PatientID,
  PatientName,
  PatientOrientation,
  PatientSex,
  PyramidUID,
  ReferringPhysicianName,
  SeriesInstanceUID,
  SeriesNumber,
  SOPInstanceUID,
  StudyDate,
  StudyID,
  StudyTime,
  Columns,
  Rows,
  TotalPixelMatrixColumns,
  TotalPixelMatrixRows,
```

Plus the hash of the compressed pixel data for the smallest pyramid level, provided it's less than 5mb.

src/openslide-vendor-dicom.c
jcupitt: ... I made `quickhash_level_pixels()` into a standard GError-style function and moved the quickhash disable into it on the "level too big" path. Open will now fail during quickhash computation for missing NumberOfFrames, failed frame open, etc.

src/openslide-vendor-dicom.c
jcupitt: I added `get_tag_int_str()`.

src/openslide-vendor-dicom.c
jcupitt: I made open fail if frame get in quickhash computation fails.

jcupitt:
OK, I think this is ready.

jcupitt:
I had another thought: maybe we should add the number of levels, and the number of associated images to the hash as well?

Imagine someone going into a DICOM directory and randomly `rm`-ing one of the levels. We'd still be able to open the file, but we probably shouldn't set the hash to be the same.

src/openslide-vendor-dicom.c
bgilbert: Thanks for the writeup @dclunie; that was very helpful!

It sounds as though the UIDs are sufficiently reliable that we can avoid hashing any pixel data, and we shouldn't need to hash a long list of attributes in order to get "enough uniqueness".  I'd particularly prefer to avoid hashing pixel data because the lowest-resolution level may not be especially small.

Since the Acquisition UID refers to a physical event rather than a particular set of output, I don't think it's helpful here.  quickhash-1 is not an integrity check, but intentional changes to slide data or metadata should change the quickhash value.  The Series Instance UID seems more useful.  I expect OpenSlide will always treat a series as indivisible; if a series e.g. has both original and derived data, OpenSlide would either ignore the extra instances or refuse to open the slide.  Thus the quickhash shouldn't need to distinguish between instances in a series.

I'm considering two options:

1. Define the quickhash to include solely the Series Instance UID (or the set of all such UIDs, once we support multi-series Z-stacks).  I have some concerns though:

   - In practice, do we need to care about the Series Instance UID collisions you mentioned?  It sounds as though those wouldn't happen under real-world conditions.

   - Is it possible for a well-behaved third-party tool to modify a series without changing the Series Instance UID?

2. Define the quickhash to include all SOP Instance UIDs.  Since the Series Instance UID is now redundant, ignore it.  This definition is more robust to alteration of the series, but possibly _too_ robust; a notional "slide" is still the same slide if someone deletes a file.  And it feels odd to avoid the UID that most directly matches our intentions.

   We'd hash all SOP Instance UIDs, not just the top (or bottom) of the pyramid, because the latter is sensitive to OpenSlide's construction of the pyramid.  We might need to change the construction logic if it turns out that we mishandle complex series.

@dclunie, does one of these options seem preferable?  Any concerns?

dclunie:
What is the "contract" for the quick hash? I.e., under what real-world scenarios is it expected to be the same, or different? @jcupitt your mention of removing files in a folder raises the possibility that this is not so well defined. Any count (number of levels or files) is presumably redundant with including all their SOP instance UIDs, if that's what you ended up doing (including those of the label and macro, if present).

bgilbert:
The quickhash should be distinct for slides with different pixels, pixel encoding, or metadata, and should be quick to calculate.  It should detect alterations performed by well-behaved software (by including top-level metadata that should always change on modification, such as unique identifiers and timestamps); need not detect arbitrary modifications such as data corruption; and should be insensitive to filesystem-level changes (e.g. file renames) that don't affect the slide semantics.  It provides a unique name for a slide, but not an integrity check.

The problem is that DICOM presents two novel issues.

1. In every other multi-file format we support, there's an index which points to the other files.  We can hash the index, and then error out if a file is missing at runtime.  But with DICOM, we don't know whether files are missing.  It isn't 100% clear whether we should think of deleting a file as creating a semantically distinct slide, or corrupting the existing one.  (I'm leaning toward the latter.)
2. In our other formats, there are clear boundaries between one "whole slide image" (as OpenSlide understands that concept) and the next.  The DICOM information model doesn't have an entity that is a perfect fit: slides can span multiple series, and acquisitions and series can both potentially include re-renderings of the same data.

Proposal 1 in https://github.com/openslide/openslide/pull/457#discussion_r1278406345 tries to map OpenSlide's concept of a "whole slide image" onto DICOM by deferring to series UIDs as stable identifiers, and considers missing instances to be corruption.  Proposal 2 makes OpenSlide responsible for the definition of "whole slide image", and treats missing instances as changing the identity of the slide.  I'd prefer proposal 1 if it's generally safe, e.g. against UID reuse and plausible user behavior, but I can't evaluate whether it is.

dclunie:
I don't have any problem with you just using the Series Instance UID, as long as the limitation is documented.

The most common reason for changing the content of a DICOM series of images stored in files in a research context would be to de-identify them by removing/changing data elements in the header and removing/replacing labels and macro images - usually people would also change the Series Instance UID (and all the other transient UIDs) in that case, but not always. The pixel data in the pyramid images would likely remain the same in such cases (though rarely burned in text like pen marks might be removed).

In a production acquisition workflow, one would not normally add to an existing series (i.e., each new scan, including a re-scan fo the same slide, gets a new Series Instance UID).

What do you do for other formats, which don't have identifiers? Actually hash the pixel data as you suggested in the beginning?

bgilbert:
Sounds good, thanks for the info.  We usually hash the lowest-resolution pyramid level plus a selection of slide metadata.  For TIFF, the latter includes the ImageDescription, which picks up XML metadata or similar.  (I just realized we don't hash XMLPacket, which would have been useful for Ventana slides.)

Hashing the lowest-resolution level is okay for proprietary formats with well-behaved pyramids, but seems iffier for interoperable formats.  We've had problems with generic TIFFs with arbitrarily large top levels.

PR #458, opened 2023-06-09
Add most dicom tags as properties

jcupitt:
Properties are named as eg.:

```
dicom.SharedFunctionalGroupsSequence.OpticalPathIdentificationSequence.OpticalPathIdentifier: '0'
```

We don't attach PerFrameFunctionalGroupSequence,
ReferenceImageNavigationSequence or PixelData, since they are huge and only relevant to DICOM image representation internals.

Binary values are NOT attached and appear as eg.

```
dicom.AcquisitionContextSequence.ConceptNameCodeSequence.LongCodeValue: '<binary value of 24 bytes>'
```

Since they can be (potentially) many megabytes in size.

### DIFF
--- a/src/openslide-vendor-dicom.c
+++ b/src/openslide-vendor-dicom.c
@@ -51,6 +51,7 @@ struct dicom_file {
 
   GMutex lock;
   DcmFilehandle *filehandle;
+  DcmDataSet *file_meta;
   DcmDataSet *metadata;
   enum image_format format;
 };
@@ -142,6 +143,7 @@ static void print_file(struct dicom_file *f G_GNUC_UNUSED) {
   debug("file:" );
   debug("  filename = %s", f->filename);
   debug("  filehandle = %p", f->filehandle);
+  debug("  file_meta = %p", f->file_meta);
   debug("  metadata = %p", f->metadata);
 }
 
@@ -205,6 +207,7 @@ static bool dicom_detect(const char *filename,
 
 static void dicom_file_destroy(struct dicom_file *f) {
   dcm_filehandle_destroy(f->filehandle);
+  dcm_dataset_destroy(f->file_meta);
   dcm_dataset_destroy(f->metadata);
   g_mutex_clear(&f->lock);
   g_free(f->filename);
@@ -317,15 +320,14 @@ static struct dicom_file *dicom_file_new(const char *filename, GError **err) {
   g_mutex_init(&f->lock);
 
   DcmError *dcm_error = NULL;
-  g_autoptr(DcmDataSet) meta =
-    dcm_filehandle_read_file_meta(&dcm_error, f->filehandle);
-  if (!meta) {
+  f->file_meta = dcm_filehandle_read_file_meta(&dcm_error, f->filehandle);
+  if (!f->file_meta) {
     _openslide_dicom_propagate_error(err, dcm_error);
     return NULL;
   }
 
   const char *sop;
-  if (!get_tag_str(meta, MediaStorageSOPClassUID, 0, &sop) ||
+  if (!get_tag_str(f->file_meta, MediaStorageSOPClassUID, 0, &sop) ||
       !g_str_equal(sop, VLWholeSlideMicroscopyImageStorage)) {
     g_set_error(err, OPENSLIDE_ERROR, OPENSLIDE_ERROR_FAILED,
                 "Not a WSI DICOM");
@@ -698,6 +700,116 @@ static bool maybe_add_file(openslide_t *osr,
   }
 }
 
+struct property_iterate {
+  openslide_t *osr;
+  const char *prefix;
+  bool first;
+};
+
+static bool add_properties_element(const DcmElement *element,
+                                   void *client);
+
+static bool add_properties_dataset(const DcmDataSet *dataset,
+                                   uint32_t index,
+                                   void *client) {
+  struct property_iterate *iter = (struct property_iterate *) client;
+  g_autofree char *new_prefix = iter->first ?
+      g_strdup(iter->prefix) :
+      g_strdup_printf("%s[%u]", iter->prefix, index);
+  struct property_iterate new_iter = { iter->osr, new_prefix, false };
+  return dcm_dataset_foreach(dataset, add_properties_element, &new_iter);
+}
+
+static char *get_element_value_as_string(const DcmElement *element, int index) {
+  DcmVR vr = dcm_element_get_vr(element);
+  DcmVRClass klass = dcm_dict_vr_class(vr);
+
+  const char *str;
+  double d;
+  int64_t i64;
+
+  switch (klass) {
+  case DCM_CLASS_STRING_MULTI:
+  case DCM_CLASS_STRING_SINGLE:
+    if (dcm_element_get_value_string(NULL, element, index, &str)) {
+      return g_strdup(str);
+    }
+    break;
+
+  case DCM_CLASS_NUMERIC_DECIMAL:
+    if (dcm_element_get_value_decimal(NULL, element, index, &d)) {
+      return _openslide_format_double(d);
+    }
+    break;
+
+  case DCM_CLASS_NUMERIC_INTEGER:
+    if (dcm_element_get_value_integer(NULL, element, index, &i64)) {
+      if (vr == DCM_VR_UV) {
+        return g_strdup_printf("%"PRIu64, i64);
+      } else {
+        return g_strdup_printf("%"PRId64, i64);
+      }
+    }
+    break;
+
+  case DCM_CLASS_BINARY:
+  default:
+    break;
+  }
+
+  return NULL;
+}
+
+static bool add_properties_element(const DcmElement *element,
+                                   void *client) {
+  struct property_iterate *iter = (struct property_iterate *) client;
+  DcmVR vr = dcm_element_get_vr(element);
+  uint32_t tag = dcm_element_get_tag(element);
+  const char *keyword = dcm_dict_keyword_from_tag(tag);
+  DcmVRClass klass = dcm_dict_vr_class(vr);
+
+  // ignore unknown tags
+  if (!keyword) {
+    return true;
+  }
+
+  if (klass == DCM_CLASS_SEQUENCE) {
+    DcmSequence *seq;
+    if (dcm_element_get_value_sequence(NULL, element, &seq)) {
+      g_autofree char *new_prefix = g_strdup_printf("%s.%s",
+                                                    iter->prefix,
+                                                    keyword);
+      struct property_iterate new_iter = { iter->osr, new_prefix, false };
+      dcm_sequence_foreach(seq, add_properties_dataset, &new_iter);
+    }
+  } else {
+    uint32_t vm = dcm_element_get_vm(element);
+
+    if (vm == 1) {
+      char *value = get_element_value_as_string(element, 0);
+      if (value) {
+        g_hash_table_insert(iter->osr->properties,
+                            g_strdup_printf("%s.%s", iter->prefix, keyword),
+                            value);
+      }
+    } else {
+      for (uint32_t index = 0; index < vm; index++) {
+        char *value = get_element_value_as_string(element, index);
+        if (value) {
+          g_hash_table_insert(iter->osr->properties,
+                              g_strdup_printf("%s.%s[%u]", 
+                                              iter->prefix, 
+                                              keyword,
+                                              index),
+                              value);
+        }
+      }
+    }
+  }
+
+  return true;
+}
+
 static void add_properties(openslide_t *osr, struct dicom_level *level0) {
   // pixel spacing is in mm, so convert to microns
   if (level0->pixel_spacing_x && level0->pixel_spacing_y) {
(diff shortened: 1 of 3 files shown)


jcupitt:
... I added the `file_meta` tags too.

src/openslide-vendor-dicom.c
```diff
@@ -698,6 +700,100 @@ static bool maybe_add_file(openslide_t *osr,
...
+  return dcm_dataset_foreach(dataset, add_properties_element, client);
+}
+
+static bool add_properties_element(const DcmElement *element,
+                                   void *client) {
+  struct property_iterate *iter = (struct property_iterate *) client;
+  DcmVR vr = dcm_element_get_vr(element);
+  uint32_t tag = dcm_element_get_tag(element);
+  const char *keyword = dcm_dict_keyword_from_tag(tag);
+  DcmVRClass klass = dcm_dict_vr_class(vr);
```
bgilbert, line 722: I don't feel strongly about it, but note that `class` isn't a reserved word in C.

src/openslide-vendor-dicom.c
```diff
@@ -698,6 +700,100 @@ static bool maybe_add_file(openslide_t *osr,
...
+
+  // ignore unknown tags
+  if (!keyword) {
+    return true;
+  }
+
+  if (klass == DCM_CLASS_SEQUENCE) {
+    DcmSequence *seq;
+    if (dcm_element_get_value_sequence(NULL, element, &seq)) {
+      char *new_prefix = g_strdup_printf("%s%s.", iter->prefix, keyword);
```
bgilbert, line 732: `g_autofree`

src/openslide-vendor-dicom.c
```diff
@@ -698,6 +700,100 @@ static bool maybe_add_file(openslide_t *osr,
...
+    char *new_value;
+
+    switch (klass) {
+    case DCM_CLASS_STRING_MULTI:
+    case DCM_CLASS_STRING_SINGLE:
+      for (uint32_t i = 0; i < dcm_element_get_vm(element); i++) {
+        const char *str;
+        if (dcm_element_get_value_string(NULL, element, i, &str)) {
+          const char *format = i > 0 ? "%s, %s" : "%s%s";
+          new_value = g_strdup_printf(format, value, str);
```
bgilbert, line 748: `GString` can be used to build up strings incrementally.  But also, while I see the compactness value of combining multiple elements in a property, we generally don't do this.  Instead, let's have each element as a separate property `foo[n]`.

src/openslide-vendor-dicom.c
```diff
@@ -698,6 +700,100 @@ static bool maybe_add_file(openslide_t *osr,
...
+            new_value = g_strdup_printf(format, value, i64);
+            g_free(value);
+            value = new_value;
+          }
+        }
+      }
+      break;
+
+    case DCM_CLASS_BINARY:
+      new_value = g_strdup_printf("%s<binary value of %d bytes>",
```
bgilbert, line 778: Let's drop this case for now.  We can always add base64 properties later if need be, but "this attribute exists and we won't tell you the contents" doesn't seem useful.

src/openslide-vendor-dicom.c
```diff
@@ -698,6 +700,100 @@ static bool maybe_add_file(openslide_t *osr,
...
+                                  dcm_element_get_length(element));
+      g_free(value);
+      value = new_value;
+      break;
+
+    default:
+      break;
+    }
+
+    g_hash_table_insert(iter->osr->properties,
```
bgilbert, line 789: Is it physically possible for a file to have multiple properties with the same key?  If so, is libdicom guaranteed to give them to us in a consistent order?

src/openslide-vendor-dicom.c
```diff
@@ -698,6 +700,100 @@ static bool maybe_add_file(openslide_t *osr,
...
+  if (klass == DCM_CLASS_SEQUENCE) {
+    DcmSequence *seq;
+    if (dcm_element_get_value_sequence(NULL, element, &seq)) {
+      char *new_prefix = g_strdup_printf("%s%s.", iter->prefix, keyword);
+      struct property_iterate new_iter = { iter->osr, new_prefix };
+      dcm_sequence_foreach(seq, add_properties_dataset, &new_iter);
+      g_free(new_prefix);
+    }
+  } else {
+    char *value = g_strdup("");
```
bgilbert, line 738: Without the inline lists, we should be able to skip the preallocation here.  If we don't recognize the type, I suspect we should skip adding the property rather than adding with an empty value.

bgilbert (commented):
I like the output from this!

> We don't attach PerFrameFunctionalGroupSequence, ReferenceImageNavigationSequence or PixelData, since they are huge and only relevant to DICOM image representation internals.

Is that filtering happening explicitly in libdicom, or is it a consequence of their types?

jcupitt:
> Is that filtering happening explicitly in libdicom, or is it a consequence of their types?

`dcm_filehandle_read_metadata()` only loads the important tags. The huge and ugly frame index, pixel data and navigation trees areas are loaded (or skipped) lazilly on frame get.

src/openslide-vendor-dicom.c
jcupitt: I've had many problems with clang interpreting C++ reserved words in plain C mode :( One of the most annoying was `or`, there's an open clang bug on this right now.

It's safest to avoid them I think.

src/openslide-vendor-dicom.c
jcupitt: Done.

src/openslide-vendor-dicom.c
jcupitt: I suppose I was imagining things like ICC profiles, where knowing there's a profile present in the metadata is useful even if you don't have the bytes available through the property system.

src/openslide-vendor-dicom.c
jcupitt: Keys are unique (phew) and always presented in numerical order (phew). It's just like TIFF.

src/openslide-vendor-dicom.c
jcupitt: I swapped to array indexing for multivalues properties, so I think this is resolved.

src/openslide-vendor-dicom.c
jcupitt: OK, done.

src/openslide-vendor-dicom.c
jcupitt: .... I've removed this.

src/openslide-vendor-dicom.c
jcupitt: Ah, it was clang-format rather than clang, but the issue is the same:

https://github.com/llvm/llvm-project/issues/59188

jcupitt:
We now have eg.:

```
dicom.ImageOrientationSlide[5]: '0'
dicom.ImageType[0]: 'ORIGINAL'
dicom.ImageType[1]: 'PRIMARY'
dicom.ImageType[2]: 'VOLUME'
dicom.ImageType[3]: 'NONE'
dicom.ImagedVolumeDepth: '0'
```

Should there also be a field for `dicom.ImageType-count`?

jcupitt:
I had to add extra indexes for sequence items, so it's now eg.:

```
dicom[0].NumberOfFrames: '139590'
dicom[0].OpticalPathSequence[0].IlluminationColorCodeSequence[0].CodeMeaning: 'Full Spectrum'
dicom[0].OpticalPathSequence[0].IlluminationColorCodeSequence[0].CodeValue: '414298005'
dicom[0].OpticalPathSequence[0].IlluminationColorCodeSequence[0].CodingSchemeDesignator: 'SCT'
dicom[0].OpticalPathSequence[0].IlluminationTypeCodeSequence[0].CodeMeaning: 'Brightfield illumination'
dicom[0].OpticalPathSequence[0].IlluminationTypeCodeSequence[0].CodeValue: '111744'
dicom[0].OpticalPathSequence[0].IlluminationTypeCodeSequence[0].CodingSchemeDesignator: 'DCM'
dicom[0].OpticalPathSequence[0].IlluminatorTypeCodeSequence[0].CodeMeaning: 'Light emitting diode'
dicom[0].OpticalPathSequence[0].IlluminatorTypeCodeSequence[0].CodeValue: '445683001'
dicom[0].OpticalPathSequence[0].IlluminatorTypeCodeSequence[0].CodingSchemeDesignator: 'SCT'
dicom[0].OpticalPathSequence[0].LensesCodeSequence[0].CodeMeaning: 'High power non-immersion lens'
dicom[0].OpticalPathSequence[0].LensesCodeSequence[0].CodeValue: '445621001'
dicom[0].OpticalPathSequence[0].LensesCodeSequence[0].CodingSchemeDesignator: 'SCT'
dicom[0].OpticalPathSequence[0].ObjectiveLensPower: '40'
dicom[0].OpticalPathSequence[0].OpticalPathIdentifier: '0'
dicom[0].PatientBirthDate: ''
```

etc., which is a bit ugly, but I think unavoidable.

jcupitt:
Sorry, once more thing, `openslide-show-properties` always sorts properties alphabetically, which is arguably wrong for dicom. Someone who knows DICOM well might be puzzled to see tags not being presented in tag order.

Perhaps properties should only be sorted alphabetically by their first element, but in application order for subelements? That would leave all the DICOM tags together and preserve the expected openslide ordering, but leave DICOM tags in DICOM order.

Just a thought, not very important or urgent.

bgilbert:
> `dcm_filehandle_read_metadata()` only loads the important tags. The huge and ugly frame index, pixel data and navigation areas are loaded (or skipped) lazilly on frame get.

Hmm, is that explicitly in the API contract?  `usage.rst` says:

> The metadata of the Data Set (i.e., all Data Elements with the exception of the Pixel Data Element and Per Frame Functional Group) can be read via `dcm_filehandle_read_metadata()`.

and the doc comment in `dicom.h` doesn't say anything about exclusions.  Asking because I'd like to make sure that down the road we don't accidentally start dumping large amounts of stuff into slide properties.

> Should there also be a field for `dicom.ImageType-count`?

Good question.  `openslide.level-count` exists, but otherwise it looks like we generally don't do this.  I'm inclined to say that we should omit for now, but I could be persuaded otherwise.

> `openslide-show-properties` always sorts properties alphabetically, which is arguably wrong for dicom. Someone who knows DICOM well might be puzzled to see tags not being presented in tag order.

We don't sort TIFF tags by tag number either.  Does the alphabetic sort order change the semantics, or just make it difficult to cross-check against the tag list?  If the latter, I'm inclined to leave it as-is.  Any application using OpenSlide properties probably doesn't understand the underlying format details anyway.

jcupitt:
> Hmm, is that explicitly in the API contract?

Yes, the docs need updating, I'm about to add some options to give the caller some control over this behaviour, and replace the printing API.

On the other two points, agreed, let's leave it, I was just thinking out loud.

src/openslide-vendor-dicom.c
```diff
@@ -698,6 +700,116 @@ static bool maybe_add_file(openslide_t *osr,
...
+
+    if (vm == 1) {
+      char *value = get_element_value_as_string(element, 0);
+      if (value) {
+        g_hash_table_insert(iter->osr->properties,
+                            g_strdup_printf("%s.%s", iter->prefix, keyword),
+                            value);
+      }
+    } else {
+      for (uint32_t i = 0; i < dcm_element_get_vm(element); i++) {
```
bgilbert, line 796:
```suggestion
      for (uint32_t i = 0; i < vm; i++) {
```

src/openslide-vendor-dicom.c
```diff
@@ -698,6 +700,116 @@ static bool maybe_add_file(openslide_t *osr,
...
+        char *value = get_element_value_as_string(element, i);
+        if (value) {
+          g_hash_table_insert(iter->osr->properties,
+                              g_strdup_printf("%s.%s[%d]",
```
bgilbert, line 800:
```suggestion
                              g_strdup_printf("%s.%s[%u]", 
```

bgilbert (commented):
Looks good!  Just some small fixes.

src/openslide-vendor-dicom.c
```diff
@@ -698,6 +700,116 @@ static bool maybe_add_file(openslide_t *osr,
...
+  struct property_iterate *iter = (struct property_iterate *) client;
+  g_autofree char *new_prefix = g_strdup_printf("%s[%u]",
+                                                iter->prefix,
+                                                index);
+  struct property_iterate new_iter = { iter->osr, new_prefix };
+  return dcm_dataset_foreach(dataset, add_properties_element, &new_iter);
+}
+
+static char *get_element_value_as_string(const DcmElement *element, int index)
+{
```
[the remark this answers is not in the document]
jcupitt: Ooop, forgot, sorry.

src/openslide-vendor-dicom.c
```diff
@@ -698,6 +700,116 @@ static bool maybe_add_file(openslide_t *osr,
...
+  case DCM_CLASS_STRING_MULTI:
+  case DCM_CLASS_STRING_SINGLE:
+    if (dcm_element_get_value_string(NULL, element, index, &str)) {
+      return g_strdup(str);
+    }
+    break;
+
+  case DCM_CLASS_NUMERIC_DECIMAL:
+    if (dcm_element_get_value_decimal(NULL, element, index, &d)) {
+      return g_strdup_printf("%g", d);
```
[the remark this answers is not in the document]
jcupitt: Done.

src/openslide-vendor-dicom.c
jcupitt: Ooop, done.

src/openslide-vendor-dicom.c
jcupitt: Done.

bgilbert (approved):
Thanks!

PR #459, opened 2023-06-12
Add jpeg2000 decode support to dicom

jcupitt:
This PR adds support for JPEG2000-coded DICOM images.

### DIFF
--- a/src/openslide-vendor-dicom.c
+++ b/src/openslide-vendor-dicom.c
@@ -177,16 +177,10 @@ static void print_frame(DcmFrame *frame G_GNUC_UNUSED) {
 }
 
 static bool dicom_detect(const char *filename,
-                         struct _openslide_tifflike *tl,
+                         struct _openslide_tifflike *tl G_GNUC_UNUSED,
                          GError **err) {
-  // reject TIFFs
-  if (tl) {
-    g_set_error(err, OPENSLIDE_ERROR, OPENSLIDE_ERROR_FAILED,
-                "Is a TIFF file");
-    return false;
-  }
-
-  // should be able to open as a DICOM
+  // some vendors use dual-personality TIFF/DCM files, so we can't just reject
+  // tifflike files
   g_autoptr(DcmFilehandle) filehandle = _openslide_dicom_open(filename, err);
   if (!filehandle) {
     return false;
--- a/src/openslide-vendor-dicom.c
+++ b/src/openslide-vendor-dicom.c
@@ -28,6 +28,7 @@
 #include "openslide-private.h"
 #include "openslide-decode-dicom.h"
 #include "openslide-decode-jpeg.h"
+#include "openslide-decode-jp2k.h"
 #include "openslide-hash.h"
 
 #include <glib.h>
@@ -43,6 +44,8 @@
 
 enum image_format {
   FORMAT_JPEG,
+  FORMAT_JPEG2000_RGB,
+  FORMAT_JPEG2000_YCBCR,
   FORMAT_RGB,
 };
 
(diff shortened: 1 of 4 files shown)


jcupitt:
Here's how to get a sample image (thanks to David for explaining this to me):

1. Go to https://console.cloud.google.com/bigquery

2. Find a suitable slide with something like:

```sql
SELECT collection_id,SeriesInstanceUID,license_short_name
FROM `canceridc-data.idc_current.dicom_all`
WHERE TransferSyntaxUID='1.2.840.10008.1.2.4.91' and Modality='SM' and SamplesPerPixel=3
LIMIT 100
```

3. Pick one (use `SeriesInstanceUID` to identify slides) and verify the licence (CC3 in this case).

4. Now find all files for this slide with eg.:

```sql
SELECT TransferSyntaxUID,StudyInstanceUID,license_short_name,ContainerIdentifier,ImageType,NumberOfFrames,gcs_url
FROM `canceridc-data.idc_current.dicom_all`
WHERE SeriesInstanceUID='1.3.6.1.4.1.5962.99.1.1135900777.339802071.1637518473321.2.0'
LIMIT 100
```

5. Install google's gsutil, see: https://cloud.google.com/storage/docs/gsutil_install#install

6. Download all the files using the `gcs_url` with a series of copies such as:

```
$ gsutil cp gs://public-datasets-idc/b4e9cf4e-b6f6-4851-9346-9a366c28c0ae/361673a1-07ea-435b-925a-9bfa84939a57.dcm .
```

I've put a copy of the slide here for your convenience, but I can't leave it there too long, it's only a small server:

http://www.rollthepotato.net/~john/1.3.6.1.4.1.5962.99.1.1135900777.339802071.1637518473321.2.0.tar.gz

About 200mb.

Here's vipsdisp viewing that slide:

image

bgilbert:
It looks like all the slides returned by that query are CC-BY, either 3.0 or 4.0.  None of the slides in openslide-testdata currently have licenses that require attribution, so adding one of those slides would add an additional constraint to the use of the dataset.  To avoid that, is there a way to obtain a CC0-licensed sample slide instead?

bgilbert:
I've rebased this and split it into two commits.  The changes LGTM once we have a test slide and test case.

jcupitt:
Ah, so close to being a PR with no changes required (other than the test image licence)!

I'll try to find a CC0 test image, I hadn't appreciated the difficulty the attribution requirement would cause.

jcupitt:
Here are two CC0 sample jp2k dicoms:

http://www.rollthepotato.net/~john/CMU_WSI_converted.tar.gz

They were generated from the openslide dataset by David Clunie using pixelmed, and should be representative of typical usage. Both seem to work well in vipsdisp.

I've put a README in there with notes and URLs.

bgilbert:
Thanks for the slides!  I noticed that the conversions didn't create dual-personality files, and since we now handle that case it'd be good to have test slides for it.  I've used the same tool to regenerate the files, one with classic TIFF and one with BigTIFF:

```
echo '{}' > empty.json
java -cp pixelmed.jar:jakarta.json-api-1.1.6.jar:javax.json-1.1.4.jar com.pixelmed.convert.TIFFToDicom \
    empty.json CMU-1-JP2K-33005.svs CMU-1-JP2K-33005/DCM ADDTIFF USEBIGTIFF 
java -cp pixelmed.jar:jakarta.json-api-1.1.6.jar:javax.json-1.1.4.jar com.pixelmed.convert.TIFFToDicom \
    empty.json JP2K-33003-1.svs JP2K-33003-1/DCM ADDTIFF DONOTUSEBIGTIFF
```

I spot-checked the DICOM dumps and didn't see any substantive differences with the original samples, so I've [uploaded these](https://openslide.cs.cmu.edu/download/openslide-testdata/DICOM/) and added tests here.  The slides don't completely read correctly (filed as #462) but that needn't block this PR.

bgilbert (approved):

jcupitt:
Hurrah! 

Oooops on #462, but it looks simple to fix, I'll make a PR for that today.

The only dual format DICOMs I have are CC BY 3, unfortunately :( Maybe we should have a separate issue for finding a CC0 sample for this case, so it's not forgotten?

bgilbert:
> The only dual format DICOMs I have are CC BY 3, unfortunately :( Maybe we should have a separate issue for finding a CC0 sample for this case, so it's not forgotten?

Ah, as I said, `TIFFToDicom` already supports creating dual-personality files, so the samples I uploaded were regenerated with that option enabled.

PR #461, opened 2023-06-13
WiP: Add ICC profile support

jcupitt:
Implemented for DICOM and TIFF-like slides.

No tests yet.

I added some docs.

I read https://github.com/openslide/openslide/issues/139#issuecomment-1505913849 but (fwiw) preferred this API:

```C
const char *openslide_read_icc_profile(openslide_t *osr, uint64_t *len);
```

ie. a pointer to openslide's internal buffer is returned, and remains valid as long as osr is valid. I wanted to avoid forcing a malloc and copy on every read, since profiles can (very unfortunately) be several 10s of MB. A function like this can also be used for has-property, ie.:

```C
    if (openslide_read_icc_profile(osr, NULL)) {
      // a profile is present
    }
```

The internal API is:

```C
void _openslide_set_icc_profile(openslide_t *osr,
                                const char *icc_profile,
                                int64_t icc_profile_length);
```

Which can be called during slide open, perhaps just after setting the properties. This thing makes a copy of the buffer and keeps it in osr, ready to hand out on request.

Marked as a draft since this is for discussion.

### DIFF
--- a/src/openslide-decode-tifflike.c
+++ b/src/openslide-decode-tifflike.c
@@ -1132,5 +1132,15 @@ bool _openslide_tifflike_init_properties_and_hash(openslide_t *osr,
   // load TIFF properties
   store_and_hash_properties(tl, property_dir, osr, quickhash1);
 
+  // attach ICC profile, if any
+  struct tiff_item *item = get_item(tl, property_dir, TIFFTAG_ICCPROFILE);
+  if (item) {
+    const char *icc_profile = _openslide_tifflike_get_buffer(tl, property_dir, TIFFTAG_ICCPROFILE, err);
+    if (icc_profile == NULL) {
+      return false;
+    }
+    _openslide_set_icc_profile(osr, icc_profile, item->count);
+  }
+
   return true;
 }
--- a/src/openslide-private.h
+++ b/src/openslide-private.h
@@ -73,6 +73,9 @@ struct _openslide {
 
   // error handling, NULL if no error
   gpointer error; // must use g_atomic_pointer!
+
+  char *icc_profile;
+  int64_t icc_profile_length;
 };
 
 struct _openslide_level {
@@ -90,10 +93,10 @@ struct _openslide_level {
 /* the function pointer structure for backends */
 struct _openslide_ops {
   bool (*paint_region)(openslide_t *osr, cairo_t *cr,
-		       int64_t x, int64_t y,
-		       struct _openslide_level *level,
-		       int32_t w, int32_t h,
-		       GError **err);
+                       int64_t x, int64_t y,
+                       struct _openslide_level *level,
+                       int32_t w, int32_t h,
+                       GError **err);
   void (*destroy)(openslide_t *osr);
 };
 
@@ -168,6 +171,10 @@ bool _openslide_clip_tile(uint32_t *tiledata,
                           int64_t tile_w, int64_t tile_h,
                           int64_t clip_w, int64_t clip_h,
                           GError **err);
+
+void _openslide_set_icc_profile(openslide_t *osr,
+                                const char *icc_profile,
+                                int64_t icc_profile_length);
 
 
 // File handling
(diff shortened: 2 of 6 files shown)


bgilbert:
> ie. a pointer to openslide's internal buffer is returned, and remains valid as long as osr is valid. I wanted to avoid forcing a malloc and copy on every read, since profiles can (very unfortunately) be several 10s of MB.

That's exactly why https://github.com/openslide/openslide/issues/139#issuecomment-1505913849 was structured that way: we shouldn't require OpenSlide to retain the profile in memory for the lifetime of the `openslide_t`.  We should treat it similarly to an associated image or slide region, and read it on request.  The application will presumably be importing the profile into its CMS, so won't be repeatedly reading the profile from OpenSlide in any event.  The internal API would then be a callback in `struct _openslide_ops`.

@lukenovak, how's your implementation in https://github.com/openslide/openslide/issues/139#issuecomment-1563425609 coming along?  If you have code ready to go, it'd be great to start the review process on that, to avoid duplication with the DICOM work.  The associated images API could happen in a second round if needed.

jcupitt:
The largest profile I've seen is 13mb from Leica, which is annoying, but not a very significant amount of memory. 3DHistech are next on about 150kb, then everyone else is under 5kb.

If the profile is read once on slide open, you can guarantee it really is only read once, you can make the getter zero-copy, there are no issues with buffer sizing or memory management in client code, and there is no need for the client to cache the result. It's simpler and faster for everyone involved.

You could also think of it in terms of who has control over profile memory and  profile read. This design has openslide doing everything, so the API is simple and openslide can prevent copies and rereads. The design in #139 splits it, with the client managing memory and openslide managing read, so the API is more complex and difficult, and responsibility for caching and reuse is pushed to the client and will need to be reimplenented in every openslide binding.

jcupitt:
Here's an SVS in vipsdisp before this PR:

image

And after:

image

bgilbert:
While ease of use is a factor in our API design, OpenSlide is a low-level API and does not solely optimize for it.  OpenSlide is intended to scale to having many open `openslide_t` handles in parallel.  The upcoming release adds cache sharing specifically for applications where a 32 MiB tile cache per handle is too expensive.  (Note that in the PR description you argued that profiles can be several tens of MB.)

And so, we're not in the business of holding large binary blobs for applications.  Our existing APIs use `read`-style interfaces for blobs (notably, associated images) to allow OpenSlide to avoid reading them if they're not used, to avoid keeping them in memory indefinitely, and to signal that the application must manage their lifecycle.  In addition: if we did do otherwise with the ICC profile, OpenSlide's copy would be redundant, since the profile blob is useless by itself.  The copy the application is _actually_ going to use is the one it loads into its CMS, which the application will still need to lifecycle.

lukenovak:
@bgilbert I would like to make one or two adjustments to how the color profile code is accessed and read, but I can try to get things going on that for quick turnaround. I will note that as of fairly recently I am no longer with PathAI so I have less access to test slides beyond the test suite, but I believe the DP200 slide that we added to the public testing repository has an embedded profile so I can use that to make sure everything is in order.

jcupitt:
True, I overstated profile sizes above. 

Perhaps the dicom profile code in this PR could be useful to you, @lukenovak:

https://github.com/openslide/openslide/pull/461/files#diff-20e021cd0c46186975c44b4700b9cee1cbf86e768492337e621f60d7786070a8

jcupitt:
I need profile support by the end of July, so I've quicklu hacked up another PR which only loads them on demand. I'll close this PR.

@lukenovak I don't mean to tread on your toes! I'm only working on this because I have to demonstrate ICC profile support pretty soon to fulfil a commitment. Your PR should take precedence.

jetic83:
Hi @jcupitt it seems like you have a way to support ICC on svs files, but the branch and PR you is deleted. 
Will there be ICC support in the master branch and/or is there should I use your branch if I need ICC support for Leica slides?

jcupitt:
Hi @jetic83, it's been replaced by this PR: https://github.com/openslide/openslide/pull/469

lukenovak:
@jcupitt @bgilbert Apologies with the delay on this. I've been dealing with some personal issues and haven't had the time to sit down and get things up like I wanted. I lost access to the code I wrote at PathAI so I was in the process of reimplementing things which was why I couldn't put things up immediately. Will take a look at the new PR, but from everything I've seen your approach in this PR roughly mirrors mine (at least in how the profile was retrieved) minus doing it when the function was called vs. storing the profile permanently.

bgilbert:
@lukenovak No problem, thanks for the update!

I think #469 should be ready to merge soon.  Since we're pushing to get the next release out the door, I might go ahead and implement ICC support for associated images as well.  It should be a pretty straightforward extension of #469.

Issue #462, opened 2023-06-14
Some images ignored in DICOM JP2K sample slides


bgilbert:
### Operating system

Fedora 38

### Platform

x86_64

### OpenSlide version

Git main

### Slide format

DICOM

### Issue details

The JP2K [sample slides](https://openslide.cs.cmu.edu/download/openslide-testdata/DICOM/) (`CMU-1-JP2K-33005` and `JP2K-33003-1`), which were converted from Aperio slides with `com.pixelmed.convert.TIFFToDicom`, have a few issues:

1. OpenSlide ignores the highest-resolution level and all the associated images because it expects their `ImageType`s to be `ORIGINAL`, but in this case they're `DERIVED`.
1. It also ignores the Aperio thumbnail, which has an image flavor of `THUMBNAIL`.  I'm not seeing that flavor defined [here](https://dicom.innolitics.com/ciods/vl-whole-slide-microscopy-image/whole-slide-microscopy-image/00080008), though.
1. If modified to accept the overview or thumbnail images, OpenSlide fails with `Not a JPEG file: starts with 0x19 0x00`.  I haven't debugged this.

Should we fix these?  I assume it's reasonable to accept `DERIVED` images when reading converted data (especially since we silently accept only the lower-resolution levels right now).  Not sure about the others.

cc @jcupitt

jcupitt:
1. I added `"DERIVED", "PRIMARY", "VOLUME", "NONE"` as another level type, which should fix this.
2. I've had DICOMs where the thumbnail has really been the LOCALIZER, which is not really the same. But this one seems to be a useful thumbnail, you're right, so we should add it. I've included that in this PR.
3. The thumbnail is 1024 x 732 pixels, so 1024 * 732 * 3 == 22487804 byes, which is the size of the object in the pixeldata. I'd guess this was uncompressed RGB and has been tagged incorrectly as JPEG. I've mailed david to ask about this.

jcupitt:
Re. 3., this is fixed by https://github.com/openslide/openslide/pull/464

PR #463, opened 2023-06-14
Improve jp2k DICOM compatibility

jcupitt:
1. Allow DERIVED associated images and DERIVED level 0
2. Support THUMBNAIL associated images
3. Fix color channels in uncompressed RGB images

See https://github.com/openslide/openslide/issues/462

### DIFF
--- a/src/openslide-vendor-dicom.c
+++ b/src/openslide-vendor-dicom.c
@@ -91,12 +91,17 @@ struct allowed_types {
 static const char *const ORIGINAL_TYPES[] = {
   "ORIGINAL", "PRIMARY", "VOLUME", "NONE", NULL
 };
+// if the image has been re-encoded during conversion to DICOM
+static const char *const DERIVED_ORIGINAL_TYPES[] = {
+  "DERIVED", "PRIMARY", "VOLUME", "NONE", NULL
+};
 static const char *const RESAMPLED_TYPES[] = {
   "DERIVED", "PRIMARY", "VOLUME", "RESAMPLED", NULL
 };
 static const char *const *const LEVEL_TYPE_STRINGS[] = {
   ORIGINAL_TYPES,
-  RESAMPLED_TYPES
+  DERIVED_ORIGINAL_TYPES,
+  RESAMPLED_TYPES,
 };
 
 static const struct allowed_types LEVEL_TYPES = {
@@ -110,12 +115,20 @@ static const char OVERVIEW_TYPE[] = "OVERVIEW";
 static const char *const LABEL_TYPES[] = {
   "ORIGINAL", "PRIMARY", LABEL_TYPE, "NONE", NULL
 };
+static const char *const DERIVED_LABEL_TYPES[] = {
+  "DERIVED", "PRIMARY", LABEL_TYPE, "NONE", NULL
+};
 static const char *const OVERVIEW_TYPES[] = {
   "ORIGINAL", "PRIMARY", OVERVIEW_TYPE, "NONE", NULL
 };
+static const char *const DERIVED_OVERVIEW_TYPES[] = {
+  "DERIVED", "PRIMARY", OVERVIEW_TYPE, "NONE", NULL
+};
 static const char *const *const ASSOCIATED_TYPE_STRINGS[] = {
   LABEL_TYPES,
-  OVERVIEW_TYPES
+  DERIVED_LABEL_TYPES,
+  OVERVIEW_TYPES,
+  DERIVED_OVERVIEW_TYPES,
 };
 static const struct allowed_types ASSOCIATED_TYPES = {
   ASSOCIATED_TYPE_STRINGS,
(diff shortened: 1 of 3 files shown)


jcupitt:
With this PR I see:

```
$ ./driver run
aperio: OK
aperio-33003: OK
...
debug-tiles: OK
dicom-tiled-full-jp2k-rgb: extended test failed: Reading associated image "thumbnail" failed: Not a JPEG file: starts with 0x22 0x00
dicom-tiled-full-jp2k-ycbcr: extended test failed: Reading associated image "thumbnail" failed: Not a JPEG file: starts with 0x17 0x00
dicom-tiled-full-jpeg: OK
...
ventana-truncated-top-level: OK
ventana-untiled-level: OK

Failed: 2/308
```

Which seems to be caused by that test image having the uncompressed RGB thumbnail incorrectly tagged as JPEG. I've asked David for clarification on this.

src/openslide-vendor-dicom.c
```diff
@@ -88,9 +88,13 @@ static const char *const ORIGINAL_TYPES[] = {
 static const char *const RESAMPLED_TYPES[] = {
   "DERIVED", "PRIMARY", "VOLUME", "RESAMPLED", NULL
 };
+static const char *const DERIVED_ORIGINAL_TYPES[] = {
```
bgilbert, line 91: Nit: let's move this before `RESAMPLED_TYPES`, and maybe add a comment explaining when we'd see it.

src/openslide-vendor-dicom.c
```diff
@@ -101,15 +105,20 @@ static const struct allowed_types LEVEL_TYPES = {
 // the ImageTypes we allow for associated images
 static const char LABEL_TYPE[] = "LABEL";
 static const char OVERVIEW_TYPE[] = "OVERVIEW";
+static const char THUMBNAIL_TYPE[] = "THUMBNAIL";
 static const char *const LABEL_TYPES[] = {
   "ORIGINAL", "PRIMARY", LABEL_TYPE, "NONE", NULL
 };
 static const char *const OVERVIEW_TYPES[] = {
   "ORIGINAL", "PRIMARY", OVERVIEW_TYPE, "NONE", NULL
```
bgilbert, line 113: We'd also need to allow `DERIVED` for `LABEL` and `OVERVIEW` if we want to fix that part of #462.

src/openslide-vendor-dicom.c
```diff
@@ -101,15 +105,20 @@ static const struct allowed_types LEVEL_TYPES = {
...
 };
+static const char *const THUMBNAIL_TYPES[] = {
+  "DERIVED", "PRIMARY", THUMBNAIL_TYPE, "RESAMPLED", NULL
```
bgilbert, line 116: Is it possible that we'd see `ORIGINAL` thumbnails?

bgilbert (commented):
In the commit message, `ORIGINAL DERIVED` seems unclear.  Typo?

src/openslide-vendor-dicom.c
jcupitt: Done.

src/openslide-vendor-dicom.c
jcupitt: I don't think so -- the thumbnail is generally the lowest pyr level, so it will have been resampled.

jcupitt:
I've fixed the RGB ordering in uncompressed RGB, or at least the `CMU-1-JP2K-33005` and `JP2K-33003-1` samples both work now.

src/openslide-vendor-dicom.c
jcupitt: Ah that's a good correction. Yes, `LOCALIZER` was bafflingly useless. I've added DERIVED too.

src/openslide-vendor-dicom.c
```diff
@@ -398,7 +421,7 @@ static void rgb_to_cairo(const uint8_t *rgb, uint32_t *dest,
                          int64_t width, int64_t height) {
   int64_t n_pixels = width * height;
   for (int64_t i = 0; i < n_pixels; i++) {
-    dest[i] = 0xff000000 | rgb[0] << 16 | rgb[1] << 8 | rgb[2];
+    dest[i] = 0xff000000 | rgb[2] << 16 | rgb[0] << 8 | rgb[1];
```
bgilbert, line 424: I agree that this makes the images render correctly, but GBR is a weird channel order, and is not how `PlanarConfiguration` 0 is documented.  Can we try to track down the underlying cause?  Is it possible that there's an off-by-one error somewhere?

jcupitt: I agree it does look odd. 

I'm reworking pixeldata parsing right now, so let's circle back to this once that is merged.

jcupitt: OK, I think the new pixeldata code has fixed this.

bgilbert:
Looks like the updated libdicom is failing the build:

```
../subprojects/libdicom/src/dicom-file.c: In function ‘print_pixeldata_create’:
../subprojects/libdicom/src/dicom-file.c:1238:36: error: parameter name omitted
 1238 | static bool print_pixeldata_create(DcmError **,
      |                                    ^~~~~~~~~~~
```

bgilbert:
I've split up the commits.

bgilbert (approved):

PR #465, opened 2023-06-24
Add `slidetool` command that subsumes existing command-line tools

bgilbert:
Add a new `slidetool` command with subcommands as follows:

- `slidetool assoc icc read` - write an associated image ICC profile to a file or stdout
- `slidetool assoc list` - list associated images
- `slidetool assoc read` - write an associated image to a PNG
- `slidetool icc read` - write an ICC profile to a file or stdout
- `slidetool prop get` - get a single property
- `slidetool prop list` - same as `openslide-show-properties`, but with a `--names` option that lists property names only
- `slidetool quickhash1` - same as `openslide-quickhash1sum`
- `slidetool region read` - same as `openslide-write-png`, but with support for writing to stdout

Include ICC profiles in PNGs produced by `assoc read`, `region read`, and `openslide-write-png`.

Make `slidetool` a multicall binary that implements the `openslide-*` commands if called by their names.  On platforms where we can (non-Windows with new enough Meson), install the `openslide-*` commands as symlinks to `slidetool`.  On other platforms, build and install redundant binaries.  The idea is for openslide-winbuild to _only_ ship slidetool, saving a few MB in the binary ZIP.

Alternative to #445.


bgilbert:
I've shorted the `associated` subcommand to `assoc` and the `property` subcommand to `prop`.

PR #466, opened 2023-06-25
Update to libdicom 1.0

jcupitt:
Update for minor API changes in libdicom 1.0.

### DIFF
--- a/src/openslide-vendor-dicom.c
+++ b/src/openslide-vendor-dicom.c
@@ -53,8 +53,8 @@ struct dicom_file {
 
   GMutex lock;
   DcmFilehandle *filehandle;
-  DcmDataSet *file_meta;
-  DcmDataSet *metadata;
+  const DcmDataSet *file_meta;
+  const DcmDataSet *metadata;
   enum image_format format;
   enum _openslide_jp2k_colorspace jp2k_colorspace;
 };
@@ -233,9 +233,7 @@ static bool dicom_detect(const char *filename,
   }
 
   DcmError *dcm_error = NULL;
-  g_autoptr(DcmDataSet) meta =
-    dcm_filehandle_read_file_meta(&dcm_error, filehandle);
-  if (!meta) {
+  if (!dcm_filehandle_get_file_meta(&dcm_error, filehandle)) {
     _openslide_dicom_propagate_error(err, dcm_error);
     return false;
   }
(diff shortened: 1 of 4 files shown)


src/openslide-vendor-dicom.c
```diff
@@ -233,8 +233,7 @@ static bool dicom_detect(const char *filename,
   }
 
   DcmError *dcm_error = NULL;
-  g_autoptr(DcmDataSet) meta =
-    dcm_filehandle_read_file_meta(&dcm_error, filehandle);
+  const DcmDataSet *meta = dcm_filehandle_get_file_meta(&dcm_error, filehandle);
```
bgilbert, line 236: Do we still need this call at all?  We don't use the value, and if `_openslide_dicom_open` now does enough to verify that this is a valid DICOM file, we could drop it.

The `G_DEFINE_AUTOPTR_CLEANUP_FUNC` for `DcmDataSet` is now unused and can be dropped.

src/openslide-vendor-synthetic.c
```diff
@@ -117,15 +117,6 @@ static bool decode_dicom(const void *data, uint32_t len,
     return false;
   }
 
-  // read Data Set metadata and frame
```
bgilbert, line 120: This line of the comment should be updated, not dropped.

src/openslide-vendor-dicom.c
jcupitt: I think we still need it -- `dcm_filehandle_create()` only checks that it can open the file. All the "is this a DICOM?" tests are in `dcm_filehandle_get_file_meta()`.

Perhaps `dcm_filehandle_create()` should do a few checks too?

src/openslide-vendor-synthetic.c
jcupitt: The comment could be eg. `// read frame`, but that's the name of the function call, so I wasn't sure it added anything. 

I think the comment was useful when you had to read the metadata before you could read the frame, but that's automatic now with libdicom 1.0, so I thought we could drop it.

src/openslide-vendor-dicom.c
bgilbert: Ah, okay.  I don't have an opinion about where libdicom should do the checks, but as it stands, makes sense to keep the call as is.

src/openslide-vendor-synthetic.c
bgilbert: Yes, except that this function has a comment block at the top of every section, so I'd like to keep that comment to keep the structure skimmable.

bgilbert:
Pushed some minor updates.

bgilbert (approved):

Issue #468, opened 2023-07-01
DICOM clone layers bug


CGDogan:
### Operating system

macOS 13

### Platform

ARM

### OpenSlide version

head of 1 Jul 2023; commit af4878bn

### Slide format

dcm

### Issue details

We have this code:

```
  for (int32_t i = 0; i < openslide_levels; i++) {
    openslide_get_level_dimensions(osr, i, &ww, &hh);
    cout << ww << " " << hh << endl;
```

Tiff reading through openslide works for us, and this code prints:

```
ca-iip	| ww,hh:
ca-iip	| 2560 1664
ca-iip	| 1280 832
ca-iip	| 640 416
ca-iip	| 320 208
ca-iip	| 160 104
```

but when I converted it to DICOM using Google's wsi-to-dicom, I get:

```
ca-iip	| 2560 1664
ca-iip	| 2560 1664
ca-iip	| 2560 1664
ca-iip	| 1280 832
ca-iip	| 640 416
ca-iip	| 320 208
ca-iip	| 160 104
```

and this explains why we get zoomed-in tiles; openslide doesn't zoom in but clones the image and adds them to the front. But is this expected behavior? I get the same quirk for all WSI DICOM files.

img1.png is the original file

Pyvips generated from [img1.png](https://github.com/openslide/openslide/assets/126820728/7bf24736-e3ee-4d49-883c-78b910458e06), img1_png_conv.tif
[img1_png_conv.tif.zip](https://github.com/openslide/openslide/files/11926881/img1_png_conv.tif.zip) and using google's wsi-to-dicom, 
[DCM_3.dcm.zip](https://github.com/openslide/openslide/files/11926883/DCM_3.dcm.zip). And I verified that we call openslide get level functions with the same parameters and the results for the dcm and tiff can be found in these folders.
[dcm3.zip](https://github.com/openslide/openslide/files/11926885/dcm3.zip)
[tiff.zip](https://github.com/openslide/openslide/files/11926886/tiff.zip)
I see that our internal zoom measure of 11 for DeepZoom in tiff gives us the same real zoom level as 9 in dicom and I suspect that off-by-two can also be seen in the code above.

But an easier testcase to help you debug is [sm_image.dcm](https://github.com/ImagingDataCommons/libdicom/blob/main/data/test_files/sm_image.dcm) which prints:

```
ca-iip	| ww,hh:
ca-iip	| 50 50
ca-iip	| 50 50
ca-iip	| 50 50
```

[...]

PR #470, opened 2023-07-08
Prevent duplicate DICOM levels

jcupitt:
DICOM directories can often have duplicate levels, perhaps if they have been copied carelessly. This PR tests for duplicate levels during dicom_open().

See https://github.com/openslide/openslide/issues/468

### DIFF
--- a/src/openslide-vendor-dicom.c
+++ b/src/openslide-vendor-dicom.c
@@ -154,27 +154,28 @@ static const struct allowed_types ASSOCIATED_TYPES = {
 
 /* The DICOM UIDs and fields we check.
  */
-static const char MediaStorageSOPClassUID[] = "MediaStorageSOPClassUID";
-static const char VLWholeSlideMicroscopyImageStorage[] =
-  "1.2.840.10008.5.1.4.1.1.77.1.6";
-static const char ImageType[] = "ImageType";
-static const char SeriesInstanceUID[] = "SeriesInstanceUID";
-static const char TotalPixelMatrixColumns[] = "TotalPixelMatrixColumns";
-static const char TotalPixelMatrixRows[] = "TotalPixelMatrixRows";
-static const char Columns[] = "Columns";
-static const char Rows[] = "Rows";
-static const char SharedFunctionalGroupsSequence[] = "SharedFunctionalGroupsSequence";
-static const char PixelMeasuresSequence[] = "PixelMeasuresSequence";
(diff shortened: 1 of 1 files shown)


bgilbert:
In the case of duplicate associated images, OpenSlide does generally just pick one of the dups, though it could also make sense to fail.  For the main image pyramid, I'm not really comfortable just ignoring the duplicate.  If there are multiple legitimate images with the same size, we could make an inconsistent set of selections.  That's the sort of case where OpenSlide prefers to fail rather than deliver bad data.

I see two possible approaches.  The simplest is to always fail if we find a main image dup.  If we want to allow the copied-file case from #468, we could ignore duplicates that have the same SOP Instance UID.

jcupitt:
Yes, that's exactly what this PR does -- ignore dupes with the same `SeriesInstanceUID`.

bgilbert:
It ignores all dups with the same Series Instance UID.  I'm saying it should, at a minimum, fail if multiple files with the same Series Instance ID have different SOP Instance UIDs but the same dimensions.

jcupitt:
I'm not certain I understand. Do you think we should ensure that `SOPInstanceUID` is different for every file in a slide? I suppose we could add a hash table of UIDs and check that whenever we select a new file. 

But it's hard to imagine a case when this test would ever fail. Is this really a useful thing to check? I'm probably missing something.

bgilbert:
This PR currently ignores a duplicate image in two cases:

1. The user has duplicated a file.  The duplicate copy would share a SOP Instance UID with the original.
2. The slide actually includes two different images with the same resolution, or two different associated images of the same type.  These would naturally have different SOP Instance UIDs.

For case 1, we could choose to ignore the duplicate, or to fail.  For case 2 we must fail, since otherwise OpenSlide's output would vary depending on the order in which the filesystem happened to return directory entries.

So, there are two possible fixes:

1. Decide that the user must not put duplicate DICOM files in the same directory.  If we find any duplicate level resolution or associated image name, fail.
2. Decide that the user may put duplicate DICOM files in a directory.  If we find a duplicate, we need to check the SOP Instance UIDs of the duplicate files.  If they match, it's because the user copied the file, and we can ignore one copy.  If they don't match, there's something about the structure of the slide that we don't understand, and we must fail.

jcupitt:
I added a thing to check SOPs, and took the opportunity to sort the DICOM keywords we use alphabetically.

bgilbert (approved):

PR #476, opened 2023-07-15
Add properties for associated image width, height, and ICC profile

bgilbert:
Associated images currently aren't represented in properties at all, which isn't entirely helpful for debugging.  Add the width, height, and ICC profile size of each associated image in the form `openslide.associated.$name.{width,height,icc-size}`.

For the prefix, I also considered `openslide.associated-image`.  That would be consistent with OpenSlide's standard terminology, e.g. in C API function names, but in a property listing it'd make the names longer (and arguably less readable) for little actual value.  `associated.label` could be read as "the associated label" so there's some logic to the shorter names.

For the image names I also considered `openslide.associated[$name]`, `openslide.associated["$name"]`, and `openslide.$name`.  We've only used brackets for array indexes, and anyway the set of image names is controlled by OpenSlide, so it seemed cleanest to use a separate hierarchy element for the name.  `openslide.$name` is compelling, but makes it harder to tell which properties correspond to associated images; namespacing them makes e.g. `grep ^openslide.associated` possible.

As with the other templated properties, leave them undocumented for now.


bgilbert:
I decided that `openslide.assoc` was too abbreviated, and switched the property prefix to `openslide.associated`.

PR #506, opened 2023-10-01
Ensure that the RT_MANIFEST value is quoted

kleisauke:
GNU windres does not treat hyphens as separators. However, llvm-rc or MSVC's rc.exe treats them as separators, causing a build error on these toolchains if the value is not quoted.

See: https://github.com/mstorsjo/llvm-mingw/issues/67


bgilbert (approved):
Thanks!

PR #515, opened 2023-10-08
openslide.h: improve wording of ICC profile docs

bgilbert:
- Avoid some circumlocution by saying "device color space"
- Hint at what the ICC profile should be used for
- Clarify "ICC profile" to "ICC color profile"


jcupitt (approved):
Look good to me!

PR #522, opened 2023-10-11
Update for release

bgilbert:


bgilbert:
The soname bump creates a circular dependency with the winbuild update.  Merging over red.

PR #558, opened 2024-02-02
dicom: set JPEG decode colorspace from DICOM metadata

jcupitt:
Hello everyone, this is a strange PR, here's the referring issue:

https://github.com/ImagingDataCommons/libdicom/issues/80

SVS doesn't set the component ids correctly in their JPEG tiles and readers are supposed to override the JPEG colorspace with the photometric interpretation of the enclosing container.  openslide does this in `decode_jpeg()` in `openslide-decode-tiff.c`.

When bioformats converts SVS to DICOM, it just copies the JPEG tiles over directly and does not fix the component ids. When openslide reads the DICOM file, libjpeg guesses YCbCr for the file encoding (it's really RGB), does the YcbCr -> RGB conversion when it should not, and you get crazy colours.

This PR adds a check for JPEG tiles with unknown component ids and only in this case overrides the `jpeg_color_space` guessed by libjpeg with the photometricinterpretation from the enclosing container.

To reproduce the issue:

```shell
bfconvert -tilex 256 -tiley 256 -compression JPEG -noflat -precompressed CMU-2.svs CMU-2.dcm
```

Then view with openslide. You'll see something like:

image

Check layer `CMU-2_0_3.dcm` (the smallest pyr layer, with 40 tiles). Extract a tile:

```shell
dcm-getframe -o x.jpg CMU-2_0_3.dcm 1
```

To make:

x

`dcm-getframe` just reads the JPEG from the file, it does no processing. If you check the component ids you'll see 0, 1, 2. It's really 4:4:4 RGB. libjpeg has this code to guess the `color_space`:

```C
    cid0 = cinfo->comp_info[0].component_id;
    cid1 = cinfo->comp_info[1].component_id;
    cid2 = cinfo->comp_info[2].component_id;

    /* First try to guess from the component IDs */
    if      (cid0 == 0x01 && cid1 == 0x02 && cid2 == 0x03)
      cinfo->jpeg_color_space = JCS_YCbCr;
    else if (cid0 == 0x01 && cid1 == 0x22 && cid2 == 0x23)
      cinfo->jpeg_color_space = JCS_BG_YCC;
    else if (cid0 == 0x52 && cid1 == 0x47 && cid2 == 0x42)
      cinfo->jpeg_color_space = JCS_RGB;    /* ASCII 'R', 'G', 'B' */
    else if (cid0 == 0x72 && cid1 == 0x67 && cid2 == 0x62)
      cinfo->jpeg_color_space = JCS_BG_RGB; /* ASCII 'r', 'g', 'b' */
    else if (cinfo->saw_JFIF_marker)
      cinfo->jpeg_color_space = JCS_YCbCr;  /* assume it's YCbCr */
    else if (cinfo->saw_Adobe_marker) {
      switch (cinfo->Adobe_transform) {
      case 0:
    cinfo->jpeg_color_space = JCS_RGB;
    break;
      case 1:
    cinfo->jpeg_color_space = JCS_YCbCr;
    break;
      default:
    WARNMS1(cinfo, JWRN_ADOBE_XFORM, cinfo->Adobe_transform);
    cinfo->jpeg_color_space = JCS_YCbCr;    /* assume it's YCbCr */
    break;
      }
    } else {
      TRACEMS3(cinfo, 1, JTRC_UNKNOWN_IDS, cid0, cid1, cid2);
      cinfo->jpeg_color_space = JCS_YCbCr;  /* assume it's YCbCr */
    }
```

So that code will guess the incorrect YCbCr for this tile, causing the incorrect colours.

If you convert the file with:

```shell
bfconvert -tilex 256 -tiley 256 -compression JPEG -noflat CMU-2.svs CMU-2.dcm
```

It works, since that will reencode all the JPEGs as 4:2:0 YCbCr and set the component ids correctly. Although the DICOM photometricinterpretation will remain RGB, which might or might not be a bug, I'm unclear.

### DIFF
--- a/src/openslide-decode-jpeg.c
+++ b/src/openslide-decode-jpeg.c
@@ -319,6 +319,7 @@ bool _openslide_jpeg_decode_buffer_dimensions(const void *buf, uint32_t len,
 
 static bool jpeg_decode(struct _openslide_file *f,  // or:
                         const void *buf, uint32_t buflen,
+                        J_COLOR_SPACE space,
                         void *dest, bool grayscale,
                         int32_t w, int32_t h,
                         GError **err) {
@@ -343,6 +344,12 @@ static bool jpeg_decode(struct _openslide_file *f,  // or:
       g_set_error(err, OPENSLIDE_ERROR, OPENSLIDE_ERROR_FAILED,
                   "Couldn't read JPEG header");
       return false;
+    }
+
+    // formats like DICOM take the colorspace from a container metadata field,
+    // not from the JPEG header
+    if (space != JCS_UNKNOWN) {
+      cinfo->jpeg_color_space = space;
     }
 
     // decompress
@@ -373,7 +380,7 @@ bool _openslide_jpeg_read(const char *filename,
     return false;
   }
 
-  return jpeg_decode(f, NULL, 0, dest, false, w, h, err);
+  return jpeg_decode(f, NULL, 0, JCS_UNKNOWN, dest, false, w, h, err);
 }
 
 bool _openslide_jpeg_decode_buffer(const void *buf, uint32_t len,
@@ -382,7 +389,17 @@ bool _openslide_jpeg_decode_buffer(const void *buf, uint32_t len,
                                    GError **err) {
   //g_debug("decode JPEG buffer: %x %u", buf, len);
 
-  return jpeg_decode(NULL, buf, len, dest, false, w, h, err);
+  return jpeg_decode(NULL, buf, len, JCS_UNKNOWN, dest, false, w, h, err);
+}
+
+bool _openslide_jpeg_decode_buffer_colorspace(const void *buf, uint32_t len,
+                                              J_COLOR_SPACE space,
+                                              uint32_t *dest,
+                                              int32_t w, int32_t h,
+                                              GError **err) {
+  //g_debug("decode JPEG buffer colorspace: %x %u", buf, len);
+
+  return jpeg_decode(NULL, buf, len, space, dest, false, w, h, err);
 }
 
 bool _openslide_jpeg_decode_buffer_gray(const void *buf, uint32_t len,
@@ -391,7 +408,7 @@ bool _openslide_jpeg_decode_buffer_gray(const void *buf, uint32_t len,
                                         GError **err) {
   //g_debug("decode grayscale JPEG buffer: %x %u", buf, len);
 
-  return jpeg_decode(NULL, buf, len, dest, true, w, h, err);
+  return jpeg_decode(NULL, buf, len, JCS_UNKNOWN, dest, true, w, h, err);
 }
 
 static bool get_associated_image_data(struct _openslide_associated_image *_img,
--- a/src/openslide-decode-jpeg.h
+++ b/src/openslide-decode-jpeg.h
@@ -48,6 +48,12 @@ bool _openslide_jpeg_decode_buffer(const void *buf, uint32_t len,
                                    int32_t w, int32_t h,
                                    GError **err);
 
+bool _openslide_jpeg_decode_buffer_colorspace(const void *buf, uint32_t len,
+                                              J_COLOR_SPACE space,
+                                              uint32_t *dest,
+                                              int32_t w, int32_t h,
+                                              GError **err);
+
 bool _openslide_jpeg_decode_buffer_gray(const void *buf, uint32_t len,
                                         uint8_t *dest,
                                         int32_t w, int32_t h,
--- a/src/openslide-vendor-dicom.c
+++ b/src/openslide-vendor-dicom.c
@@ -63,6 +63,7 @@ struct dicom_file {
   const DcmDataSet *metadata;
   const char *slide_id;
   enum image_format format;
+  J_COLOR_SPACE jpeg_colorspace;
   enum _openslide_jp2k_colorspace jp2k_colorspace;
 };
 
@@ -461,8 +462,9 @@ static bool decode_frame(struct dicom_file *file,
 
   switch (file->format) {
   case FORMAT_JPEG:
-    return _openslide_jpeg_decode_buffer(frame_value, frame_length,
-                                         dest, w, h, err);
+    return _openslide_jpeg_decode_buffer_colorspace(frame_value, frame_length,
+                                                    file->jpeg_colorspace,
+                                                    dest, w, h, err);
   case FORMAT_JPEG2000:
     return _openslide_jp2k_decode_buffer(dest, w, h,
                                          frame_value, frame_length,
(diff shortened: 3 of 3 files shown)


jcupitt:
A question I had was if the DICOM PI tag is expected to be correct. Many of these tags are there to record the history of the pixels, not their current encoding.

An alternative solution would be for `jpeg_decode()` to just test for component ids set to 0/1/2, assume that meant a broken SVS JPEG, and force `jpeg_color_space` to RGB in this case. Maybe that's safer and less intrusive?

bgilbert:
Thanks for the PR!

Actually, the Aperio SVS files are correct.  TIFF doesn't require JPEG component IDs to have any specific value.  While JFIF mandates component IDs of 1, 2, and 3 (for YCbCr images, since JFIF doesn't support RGB), TIFF doesn't require JPEGs to adhere to JFIF.  The [spec](https://web.archive.org/web/20210108174708/https://www.adobe.io/content/dam/udp/en/open/standards/tiff/TIFFphotoshop.pdf) for TIFF compression 7 (new-style JPEG) says:

> The component ID bytes are arbitrary so long as each component within an image segment is given a distinct ID.

And likewise, [the JPEG spec](https://www.w3.org/Graphics/JPEG/itu-t81.pdf) doesn't require specific values for component identifiers:

> These values shall be used in the scan headers to identify the components in the scan.  The value of C<sub>i</sub> shall be different from the values of C<sub>1</sub> through C<sub>i − 1</sub>.

On the other hand, the DICOM spec [says](https://dicom.nema.org/medical/dicom/current/output/html/part05.html#sect_8.2.1) that the photometric interpretation should be governed by the DICOM Data Element:

> When decompressing, should the characteristics explicitly specified in the compressed data stream (e.g., spatial subsampling or number of components or planar configuration) be inconsistent with those specified in the DICOM Data Elements, those explicitly specified in the compressed data stream should be used to control the decompression. [...] Those characteristics not explicitly specified in the compressed data stream (e.g., the color space of the compressed components, which is not specified in the JPEG Interchange Format) [...] can therefore be determined from the DICOM Data Element in the enclosing Data Set. For example a Photometric Interpretation of "YBR_FULL_422" would describe the color space that is commonly used to lossy compress images using JPEG. [...] The JPEG Interchange Format is distinct from the JPEG File Interchange Format (JFIF). [...] The JPEG Interchange Format does NOT specify the color space.

So we do have a bug; OpenSlide should be changed to always use the PhotometricInterpretation from the DICOM element.  That should fix the `bfconvert -precompressed` case, but will break `bfconvert` without `-precompressed`.

The solution for the latter is for `bfconvert` to correctly set the DICOM PhotometricInterpretation.  In general, OpenSlide carries workarounds for format bugs produced by slide scanners, but does not carry workarounds for format bugs produced by postprocessing tools.  That restriction helps keep the codebase manageable, since the file can be fixed by re-postprocessing the original scan.

src/openslide-decode-jpeg.c
```diff
@@ -373,16 +389,17 @@ bool _openslide_jpeg_read(const char *filename,
     return false;
   }
 
-  return jpeg_decode(f, NULL, 0, dest, false, w, h, err);
+  return jpeg_decode(f, NULL, 0, space, dest, false, w, h, err);
 }
 
 bool _openslide_jpeg_decode_buffer(const void *buf, uint32_t len,
+                                   J_COLOR_SPACE space,
```
bgilbert, line 396: Let's add a new wrapper function e.g. `_openslide_jpeg_decode_buffer_colorspace()` with the extra argument, rather than changing the existing callers to pass `JCS_UNKNOWN`.  That constant feels a bit too magic, and `_openslide_jpeg_decode_buffer_gray()` already provides precedent for using a separate function.

src/openslide-decode-jpeg.c
```diff
@@ -345,6 +346,20 @@ static bool jpeg_decode(struct _openslide_file *f,  // or:
...
+    // component_ids and we have been given a known colorspace
+    if (space != JCS_UNKNOWN && cinfo->num_components == 3) {
+      int cid0 = cinfo->comp_info[0].component_id;
+      int cid1 = cinfo->comp_info[1].component_id;
+      int cid2 = cinfo->comp_info[2].component_id;
+
+      if (cid0 == 0 && cid1 == 1 && cid2 == 2) {
+        // SVS sometimes mistakenly uses 0/1/2 to mean RGB, eg. CMU-2.svs
+        // sample file
+        cinfo->jpeg_color_space = space;
```
bgilbert, line 359: As discussed above, we should override the input space whenever it isn't specified as `JCS_UNKNOWN`.

bgilbert (commented):
Looks good!  Just a couple small changes.

src/openslide-decode-jpeg.c
jcupitt: ok, done!

src/openslide-decode-jpeg.c
jcupitt: ok, done

jcupitt:
Great! It feels good to resolve this.

However ... this will cause pain for some users until bioformats fix their converter. We could have a workaround for this case, ie. spot 0/1/2 as the component ids and force RGB in that case. It's only a few lines and would (I think?) not cause too much extra maintenance. We could maybe show a warning and urge users to investigate?

jcupitt:
I made a `bfconvert` issue: https://github.com/ome/bioformats/issues/4161

bgilbert:
Thanks for filing that issue!

> We could have a workaround for this case, ie. spot 0/1/2 as the component ids and force RGB. It's only a few lines and would (I think?) not cause too much extra maintenance.

There's no guarantee that 0/1/2 means Aperio + RGB.  Some other DICOM encoder on some platform could plausibly write YCbCr JPEG tiles (not using JFIF) with those IDs.

OpenSlide might conceivably carry such a workaround, but only if it were clear that this was a long-term problem for a significant set of files/users.  OpenSlide doesn't follow Postel's law, basically.  It's better to fail in an obvious way so problems can be found and fixed, rather than muddling on and compounding them.

jcupitt:
OK, that sounds reasonable. I'll make the `bfconvert` issue sound a bit more alarming, since this is very likely to bite people and they need to fix this. And they probably have a matching bug in their DICOM reader.

bgilbert:
For the record, even with `-precompressed`, `bfconvert` re-encodes the associated images to YCbCr but leaves their PhotometricInterpretation as RGB.  So, for Aperio slides converted in this way, this PR fixes the main pyramid but breaks associated images.  That'll likewise need to be fixed in `bfconvert`.

Thanks again!

jcupitt:
Oh nice spot, I should have thought to check the label encoding.

PR #565, opened 2024-02-23
dicom: ignore dual-personality TIFF files with `.tif`/`.tiff` file extension

bgilbert:
If the specified slide file has a TIFF filename extension, assume the user wants to read it as TIFF, and have the DICOM driver ignore it.

### DIFF
--- a/src/openslide-vendor-dicom.c
+++ b/src/openslide-vendor-dicom.c
@@ -558,11 +558,17 @@ static const struct _openslide_ops dicom_ops = {
   .destroy = destroy,
 };
 
-static bool dicom_detect(const char *filename,
-                         struct _openslide_tifflike *tl G_GNUC_UNUSED,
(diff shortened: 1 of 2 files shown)


jcupitt (commented):
Looks good here!

I agree that letting users override file type detection by modifying the filename extension could be useful.

What does openslide do with eg. an SVS renamed as TIFF? Maybe we need a parallel change in places like that to keep behaviour consistent?

bgilbert:
In general, OpenSlide uses content detection and doesn't care about file extensions.  The dual-personality DICOM-TIFF case is unusual because the file is intended to be parsable as two different formats, and the two sets of metadata may not match.  So, since it's not necessarily clear which format is intended to be primary, this PR lets the user give us a hint.  For other TIFF-based files, allowing such flexibility wouldn't be useful: if OpenSlide has a vendor-specific driver for a particular TIFF, that's because the generic TIFF driver would not produce valid and complete output for that file.

PR #600, opened 2024-05-15
support zstd compressed CZI

iewchen:
CZI has two zstd compression modes: zstd0 and zstd1.

In zstd0 mode, pixel data is compressed with zstd and stored in subblock
as it is.

zstd1 mode is different in that it prefix the zstd compressed data with
a header. This header is either 1 byte or 3 bytes long. The first byte
of the header is its length. CZI may use a trick called high low byte
unpack, which packs less significant byte of 16bits pixels in the first
half of the image array, and more significant byte in the second half of
the image array, before been compressed by zstd. This trick is used if:

- the header length is 3, and

- the second byte in the header is 1, and

- the lowest bit of the third byte is 1

Obviously this trick only applies to 16bits grayscale images and 48bits
color images.


src/openslide-vendor-zeiss.c
```diff
@@ -360,127 +360,130 @@ static void bgr48_to_argb32(uint8_t *src, size_t src_len, uint32_t *dst) {
...
+  // bytes in the second half of the image array.  Undo this.
+  g_autofree uint8_t *unhilo_data = NULL;
+  if (do_hilo) {
+    if (pixel_bytes % 2) {
+      g_set_error(err, OPENSLIDE_ERROR, OPENSLIDE_ERROR_FAILED,
+                  "Can't perform HiLo unpacking with an odd number of bytes "
+                  "%"PRId64, pixel_bytes);
+      return false;
+    }
+    int64_t half_bytes = pixel_bytes / 2;
```
iewchen, line 467: Later when add SIMD, the undo HiLo pack will be moved to an individual function, which will be dispatched to pure C or SIMD, based on detected CPU features.

iewchen:
Thank you for the review!

I tested the latest commit. It works.

bgilbert:
Great, thank you for the PR!

The next two PRs will probably take longer to land.  I assume the JXR one will be pretty straightforward to review, but it may not be able to land right away because of the libjxr situation.  The SIMD one will likely require substantial effort to review and test.  If it makes sense to submit both in parallel, feel free, and otherwise I'd suggest submitting the JXR one next.  I may not be able to review the SIMD one for a couple months or more.

PR #602, opened 2024-05-15
Large NDPI file fix

### LINKED ISSUE #174: Certain (very large) Hamamatsu NDPI files cannot be opened with OpenSlide

ntrahearn:
Addresses issue https://github.com/openslide/openslide/issues/174 (and possibly others), regarding the reading of large NDPI files. This issue stems from the fact that classical TIFF (and by extension, NDPI) images only support up to 32 bit values for tagged metadata. However, due to the nature of whole slide image data, it isn't uncommon for the size of an NDPI to exceed the 32 bit range. This means that key metadata, such as the byte positions of image layers or their size in bytes, may be too large to store in a traditional TIFF IFD entry.

Currently OpenSlide relies on heuristics to determine the high bits of 64 bit addresses, which fail in some cases. However, this is unnecessary, as NDPI actually stores the high bits of the offset/value of each tag in 4 byte blocks immediately after the end of the IFD.

This fix modifies openslide-decode-tifflike.c to append these extra 4 bytes to each IFD entry's value/offset and, if necessary, modifies its type to LONG8.

This fix also modifies openslide-vendor-hamamatsu.c to construct correct restart marker addresses. Currently only the values in TIFF tag 65426 are used, which are only the lower 32 bits of each address. High bits are stored in TIFF tag 65432, so these are now appended before mcu_starts are calculated


bgilbert:
There's a substantial backlog of PRs I need to go through; I'll get to it but not yet.  I'm working on improving the testing infrastructure first.

joss94:
Hi ! 

Thank you so much for the amazing work on OpenSlide !
Any update on when this PR would be merged ? I guess it would be  really helpful to many people, it sure would help me a lot !

Thanks again

cc-jj:
@bgilbert thank you for your work on Openslide. Can you provide an ETA for when this fix will be merged?

carlossantosgarcia:
Hello !

Thanks for your work, any updates on this?

bgilbert:
Thank you again for tracking this down and submitting a fix!  (And for your extreme patience waiting for it to merge.)

PR #610, opened 2024-05-28
misc/imhex: add Hamamatsu NDPI pattern

bgilbert:
Based on upstream ImHex TIFF pattern as of commit WerWolv/ImHex-Patterns@10bf1c76cf, licensed under GPLv2.  Support for 64-bit NDPI based on work by @ntrahearn.


jaharkes:
Wrong commit id in the PR comment? Looks like it should be 7ecd6d8

bgilbert:
Yeah, I was worried that might be unclear.  That commit corresponds to the tree I pulled the upstream file from, not the most recent commit that changed the file.

jaharkes:
Ah ok. In that case ignore me and carry on.

Issue #592, opened 2024-05-06
Support KFB


XKhelloworld:
### Slide format

KFB

### Issue details

Hello! I am a Chinese user. May I ask whether the products of your company (openslide) can support the direct operation of the.KFB format file, which is from a medical enterprise in Zhejiang Province, China? Thank you so much!!!

peterchen183:
> ### Slide format
> KFB
> 
> ### Issue details
> Hello! I am a Chinese user. May I ask whether the products of your company (openslide) can support the direct operation of the.KFB format file, which is from a medical enterprise in Zhejiang Province, China? Thank you so much!!!

Hello! Could you provide some sample KFB files, I've made some progress with supporting this format(not 100% done, but very close I think..).

peterchen183:
> Hello! OpenSlide currently does not support KFB files. We'd be interested in supporting the format if someone can contribute sample KFB files and code to support KFB.
> 
> (Sample files can be uploaded [here](https://openslide.org/submit/). Code must be developed without using any information affected by licensing restrictions or a non-disclosure agreement.)



> Hello! OpenSlide currently does not support KFB files. We'd be interested in supporting the format if someone can contribute sample KFB files and code to support KFB.
> 
> (Sample files can be uploaded [here](https://openslide.org/submit/). Code must be developed without using any information affected by licensing restrictions or a non-disclosure agreement.)

/*
 * KFBio (kfb) support
 */

#include "openslide-private.h"
#include "openslide-decode-jpeg.h"
#include "openslide-decode-tiff.h"
#include "openslide-decode-tifflike.h"

#include <glib.h>
#include <string.h>
#include <stdlib.h>
#include <tiffio.h>

static const char KFB_EXT[] = ".kfb";

struct slide_zoom_level_section {
  int concat_exponent;

  double overlap_x;
  double overlap_y;
  double mpp_x;
  double mpp_y;

  uint32_t fill_rgb;

  // enum image_format image_format;
  int image_w;
  int image_h;
};

struct slide_zoom_level_params {
  int image_concat;
  int tile_count_divisor;
  int tiles_per_image;
  int positions_per_tile;
  double tile_advance_x;
  double tile_advance_y;
};

struct image {
  int64_t start_in_file;
  int32_t length;
  int32_t imageno;   // used only for cache lookup
  int32_t iw;
  int32_t ih;
  int refcount;
};

struct tile {
  struct image *image;

  // location in the image
  double src_x;
  double src_y;
};

struct level {
[...]

PR #616, opened 2024-06-06
kfb support

peterchen183:
This is the work for supporting KFBio - KFB format.
#592


peterchen183:
Please ingore this one, will commit a signed-off PR.

PR #621, opened 2024-06-15
Emscripten build

Siirko:
There is a Python bindings, I thought it may interest some people to have wasm build.

The meson setup part is passing, but the compilation is stucked at the half of the compilation
image

this PR is for the man or peoples who will have the faith and need to have a WASM build.


bgilbert:
Thanks for the PR!  We don't include configuration for binary builds in this repo, but [openslide-bin](https://github.com/openslide/openslide-bin/) would be a great place to add this if you're interested!

PR #625, opened 2024-08-08
Add Support for InteMedic-tron

peterchen183:
This is the work for supporting InteMedic-tron format.


peterchen183:
How to add more dependencies (libssl, libcrypto and json-c), to pass all the checks?

Thank you.
Peter.

bgilbert:
Thanks for contributing additional formats to OpenSlide!  I really appreciate the work you're putting in, and it's great to have code (and, implicitly, format documentation) available in the repo for users who need to support these formats.

To answer your question, you'll need to install the additional dependencies in the [GitHub Actions workflow](https://github.com/openslide/openslide/blob/main/.github/workflows/c.yaml).  The Windows build will still not work until openslide-bin is updated; see below.

I haven't seen explicit user interest in this format, which makes it difficult to prioritize the substantial amount of work needed to land this code (or indeed any new OpenSlide vendor driver).  The best way for you to get a new format merged is to do as much of the preparatory work as possible, so that the PR is easy to review and requires minimal feedback.  In a _really_ quick skim, I noticed a variety of problems:

- Reimplementation of AES, rather than using the AES routines in libcrypto.
- A PBKDF implementation that's only called on a single constant value.  You can just hardcode the PBKDF output instead.  (Are you absolutely, 100% sure that all files use the same key?)
- A libssl dependency, which seems like it shouldn't be necessary?
- Reimplementation of a ZIP decoder, rather than using an existing library.  [libgsf](https://gitlab.gnome.org/GNOME/libgsf) is a reasonable option, though it'd need a Meson wrap for openslide-bin.
- `_2` implementations of existing functions, rather than adding the needed functionality directly to those functions.  (The tilemap `read_missing_tile` function isn't needed, since it's the application's responsibility to apply the background color when transparent pixels exist.)
- Various misuses of the [GError API](https://docs.gtk.org/glib/error-reporting.html#rules-for-use-of-gerror).
- [Manual deallocation](https://openslide.org/docs/newformat/#memory-allocation) of memory allocated within the same stack frame.
- A lot of manual I/O which will need to be carefully checked for overflows/underflows due to user-controlled data.
- Various code that doesn't conform with OpenSlide coding conventions: camelCase variables, unbraced `if` blocks, overlong lines, some `g_free(g_steal_pointer(&x))` instead of letting the destructor run normally, etc.

In order to merge this, we would also need:

- [Redistributable sample data](https://openslide.org/submit/).
- [Test cases](https://openslide.org/docs/testsuite/) &mdash; positive tests at least, and preferably also negative tests.  These can't be created until sample data has been submitted and added to openslide-testdata.
- A PR to [openslide-bin](https://github.com/openslide/openslide-bin/) that [adds the necessary dependencies](https://github.com/openslide/openslide-bin/?tab=readme-ov-file#adding-a-new-dependency) for the binary builds.
- A PR to the [website](https://github.com/openslide/openslide.github.io) adding documentation on the internals of the new format.
- [README](https://github.com/openslide/openslide/blob/main/README.md) updates to list the additional library dependencies and link to the new format documentation.
- [OpenSlide Python](https://github.com/openslide/openslide-python/) documentation updates to link to the new format documentation.

I can see that you've put a lot of effort into getting this format working in OpenSlide.  Again, thank you!  The reality is that all of the above and more will need to be addressed before this code can land, and I can spend very little time on it myself.  But if you're able to do most of the work, I'd love for OpenSlide to add support for additional formats.

peterchen183:
@bgilbert  Thank you for the detailed reply!
I will do more work to get this PR ready, so close it for now.

bgilbert:
Feel free to keep the PR open while you continue to work on it.  You can mark the PR as a draft if you'd like.

Issue #392, opened 2022-08-24
Question: Artefacts in slide.read_region at locations not divisible by 2^level


leclerc91:

bgilbert:
Artifacts are not expected, and you should be able to pass any pixel coordinates to `read_region`.  Could you test on a current distro release, maybe Ubuntu 22.04?  It's possible that this is a cairo issue, or an interaction between the newer pixman and older cairo.

leclerc91:
I have tested it in Ubuntu 22.04.1 and I got the same result :(
First, I created a new fresh `conda` environment with the following packages:
`openslide openslide-python` (from **conda-forge**)
`pixman=0.40.0` (from **anaconda**)
`jupyter matplotlib` (from **main**)

Later, I installed even the latest version of `cairo` (via `apt-get install libcairo2-dev`) and removed `cairo` from my conda environment (with `--force` to keep the other packages there). Again, same result.

Any other suggestions to try?

464hee:
![11](https://user-images.githubusercontent.com/100740023/203204728-32ed6ef8-cfbe-4c00-a492-b313f3a6ac0d.png)
I'm very lucky to meet your problem. I have also encountered this situation. At present, I plan to use half step fusion to solve it

464hee:
![EE101DD5-5BDA-4eb1-8C77-E1A1950B2D30](https://user-images.githubusercontent.com/100740023/203208751-449b4e2e-babc-4fb2-abbf-ad854eb717e5.png)
The problem still exists
failed

464hee:
@bgilbert @leclerc91 Do you have any new ideas

Issue #641, opened 2024-11-19
Avoid subpixel cairo_translate()


iewchen:
### Slide format

SVS CZI

### Issue details

There were discussions on whether the SATURATE operator is needed or can be ignored (#440). It seems the SATURATE operator was introduced to avoid seams caused by Cairo's subpixel translation. It can happen with certain combinations of x, y, and tile size, when the tiles have downsample factor greater than 1.

With subpixel translation, every pixel of an image tile is moved to a position between the pixels of the output surface. Therefore, the values of the output pixels require interpolation. This interpolation uses computing resources, makes the image appear Gaussian-smoothed, and introduces seams that require slow SATURATE operator to fix.

What if instead of doing subpixel translation, do integer-valued translation? In my tests, the less correct integer translation actually looks better. By applying `floor()` to the `tx` and `ty` values passed to `cairo_translate()`, a tile is positioned slightly less than one pixel to the left or above its subpixel position. No SATURATE operator is required. While integer translation can introduce artifact, I could not see any on the tile borders. On the other hand, the image is noticeably sharper since the tiles remain unmodified.

Consider two non-overlapping adjacent tiles, the right border of first tile is moved to `x1 = floor(d1.d2)`, where d1, d2 are the integer and decimal part of a float number. The left border of the second tile is moved to `x2 = floor(d1.d2 + 1/downsample_factor)`. Depends on the magnitude of `d2` and `1/downsample_factor`(even when there is a gap `w` bewteen two adjacent tiles on their full resolution projection, this term becomes `(w+1)/downsample_factor` and shall still less than 1), `x2` is either equals to `x1 + 1`, or `x1`. That is, the two tiles are either aligned perfectly without overlapping, or have exact one pixel overlap. Using `floor()` provides more consistent results, particularly when one of two adjacent tiles has a negative translation and the other has a positive translation. In contrast, `round()` might result in a tile being shifted one pixel more or one pixel less than it should be.

I tested on Aperio CS SVS (which uses simple grid) and Zeiss CZI (which uses range grid). After comment out `cairo_set_operator(cr, CAIRO_OPERATOR_SATURATE);` in `openslide.c`, there are seams when jump around slides randomly on levels with downsample > 1. At the locations where seams happen, pass integer-valued `tx` `ty` to `cairo_translate()` makes seams disappear.

bgilbert:
[...]

Issue #612, opened 2024-05-29
Pixels


EricRobert:
### Operating system

Ubuntu 20.04

### Platform

64-bit x86

### OpenSlide version

latest from PPA i.e. libopenslide.so.1.0.0 + openslide_python-1.3.1

### Slide format

SVS

### Issue details

I found a discrepancy when comparing the output of `read_region` of 3 libraries with SVS support where OpenSlide doesn’t match the other 2 i.e. compared to [cuCIM](https://github.com/rapidsai/cucim) and [tifffile](https://github.com/cgohlke/tifffile).

So far, the problematic pixels seems to be limited to `read_region` of non-zero layers that are large enough so that either the row or the column is >= 4096.

I used this code to compare:

```python
import numpy
from cucim.clara import CuImage
from openslide import OpenSlide
from PIL import Image
from tifffile import TiffFile

path = "/wsi/CMU-1.svs"

s = OpenSlide(path)
c = CuImage(path)
assert s.level_dimensions == c.resolutions["level_dimensions"]

for level in reversed(range(s.level_count)):
    w, h = s.level_dimensions[level]
    print(f"{w}x{h} @{level}")

    # read openslide
    a = numpy.asarray(s.read_region((0, 0), level, (w, h)))
    a = a[:, :, :3]

    # read cucim
    b = numpy.asarray(c.read_region(location=(0, 0), level=level, size=(w, h)))

    # if there is any difference, write it down
    if not numpy.array_equal(a, b):
        img = Image.fromarray(numpy.all(a == b, axis=-1))
        img.save(f"diff.{level}.png")

        # produce a few downscaled versions to make it friendly for viewers
        img = img.convert("L")
        for i in range(s.level_count - level):
            img.reduce(4 * (i + 1)).save(f"diff.{level}.{i}.png")

        # compare both against tifffile
        with TiffFile(path) as f:
            r = f.series[0].levels[level].asarray()
            r = r[:h, :w, :]
            print(numpy.array_equal(r, a), "vs. openslide")
            print(numpy.array_equal(r, b), "vs. cucim")
```

Here is an example using `CMU-1.svs` output that shows that there's no discrepancies in [(0, 0) to (4096, 4096)[ but then there's some.

```
2875x2057 @2
11500x8228 @1
False vs. openslide
True vs. cucim
46000x32914 @0
```

![diff 1 1](https://github.com/openslide/openslide/assets/1616287/88672e22-0e40-4c4a-8ce0-3f45c9770844)

[...]

Issue #552, opened 2024-01-12
openslide_read_region is slower in 4.0.0(Windows build 20231011) compared with 3.4.1(Windows build 20150527)


peterchen183:
### Operating system

Windows 10 Pro

### Platform

64-bits x86

### OpenSlide version

4.0.0

### Slide format

SVS

### Issue details

I used to use "openslide_read_region" to read from SVS (in my case the "CMU-1-Small-Region.svs"), after I updated binaries version from 3.4.1(Windows build 20150527) to 4.0.0(Windows build 20231011), I changed no code on my side, only updated binaries, I see no speedup, but x2-x3 slower.
I also tested with some other SVS and Generic Tiff, and set parameter "level" of "openslide_read_region" as 0 or 1, samer behavior.

bgilbert:
Could you try with build 20230414?  That's the most recent build with OpenSlide 3.4.1.

What parameters are you passing to `openslide_read_region()` and how are you benchmarking the function call?  What language are you calling it from?  `CMU-1-Small-Region.svs` may not be a good benchmark because it's so small.

peterchen183:
> Could you try with build 20230414? That's the most recent build with OpenSlide 3.4.1.
> 
> What parameters are you passing to `openslide_read_region()` and how are you benchmarking the function call? What language are you calling it from? `CMU-1-Small-Region.svs` may not be a good benchmark because it's so small.

Thanks for the quick reply!
Benchmark is done in C# code(call the same method with same parameters and repeat for 1000 times), I wrote a wrapper and call from C# via P/Invoke.
parameters: x:0, y:0, level:0, w:2220, h:2967
total time(1000 times) compared between different version:
20150527  70s
20160612 175s
20230414 174s
20231011 157s
testing with Image of CMU-1.svs

You can see that the difference is introduced in build 20160612, so I tried to find which updated dependencies caused this. It turns out to be libpixman-1-0.dll(pixman?).
When I update all dependencies but leave pixman alone, from build 20150527 to 20160612, the total time did not increase, still about 70s.

bgilbert:
Great, thanks for the additional info!  As an initial hypothesis, this sounds like a pixman regression, possibly specific to the `SATURATE` drawing operator (which isn't commonly used outside of OpenSlide).  I'll try to make some time to reproduce the issue and bisect pixman.

peterchen183:
[...]

PR #643, opened 2024-11-26
Use integer-valued `cairo_translate()` for performance and image quality

iewchen:
Integer-valued `cairo_translate()` is faster and has better image quality, it also avoids the slow SATURATE compositing operator. See https://github.com/openslide/openslide/issues/641 for details.

Currently, this is only enabled for Aperio, Leica, and Zeiss formats, as I have many slides in these formats and am confident this approach works correctly with them.

### DIFF
--- a/src/openslide-grid.c
+++ b/src/openslide-grid.c
@@ -200,6 +200,33 @@ static void matrix_restore(struct cairo_matrix *m) {
 typedef struct cairo_matrix cairo_matrix;
 G_DEFINE_AUTO_CLEANUP_CLEAR_FUNC(cairo_matrix, matrix_restore)
 
+/* do integer or float point translation based on Cairo compositing operator */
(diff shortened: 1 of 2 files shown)


iewchen:
I can confirm this PR fixes #392, #552 and #612. It should work for #256 and #417.

#256 Read Region without Losing Data
Level 0 x,y were not mapped to integer valued coordinates on downsampled level because Aperio's downsample values are slight off integer(e.g. 4.000152508769254), and pixel values were changed by interpolation.
    
#392 Question: Artefacts in slide.read_region at locations not divisible by 2^level
Artifacts were from unable to interpolate boolean mask.
    
#417 Inconsistent pixel value when fetching with non zero offset
Pixel values are changed by interpolation
    
#552 Very slow drawing with Pixman
Slow SATURATE operator to fix seams from subpixel translation.

#612 Decoding discrepancies in large regions
Similar to #256 and #417.

PR #648, opened 2025-01-12
Add dicom concatenations

### LINKED ISSUE #511: Multiple .dcm files per layer. OpenSlideError: Slide contains unexpected image
### Operating system

Linux Ubuntu

### Platform

x86-64

### OpenSlide version

Latest with DICOM support

### Slide format

DICOM

### Issue details

Hello,

We have issue reading some Dicom Whole Slides Images from 3DHISTEC P1000 scanners. We have noticed that a certain scanner parameter (without extended offset table enabled) caps each .dcm file to a maximum size. 
This leads to the patches of 40X or 20X magnifications to be split into multiple .dcm files. In the metadata of each file, I can find the offset and number of patches:
(0020, 9228) Concatenation Frame Offset Number
(0028, 0008) Number of Frames

When I try to read it with OpenSlide, I get the following error:
OpenSlideError: Reading slide_name.dcm: Slide contains unexpected image (1.2.752.24.16.1.2892922223.1240020768.775099804.2140302833 vs. 1.2.752.24.16.1.3390999071.1304012305.668579475.1197128784)

We have generated many files using different parameters including tiled_sparse and tiled_full. Although we do not use tiled_sparse in our scanning parameters, I thought I would also include these files in case you want to add support for them, none of them currently work and return the following message:
Detail: libdicom Parse error: Reading PixelData failed - Could not determine offset of Pixel Data Element.

Considering the default parameters for DICOM of 3DHistech scanners are tilled_full without extended offset, this might be a recuring issue also for other users, especially for slides of which the 40X layers go beyond the per file cap.

Using pydicom and reading the files patch by patch, we managed to create a tif with the Whole Slide Image which is readable on OpenSlide.

jcupitt:
Layers in large DICOM WSI files are commonly split into separate files to make transmission easier. This PR adds support for `TILED_FULL` and `TILED_SPARSE` scans with split levels. 

Referring issue:

See https://github.com/openslide/openslide/issues/511

Overview:

1. The openslide DICOM loader has separate structures for `dicom_level` and `dicom_file`, but only allows one `dicom_file` per level. This PR adds an array of `dicom_file` for each level, with an array `int file_index[n_tiles]` giving the index of the file containing each tile.
2. During directory scan, files are attached to their matching level.
3. In `read_tile()`, the `file_index` array is constructed on first use (protected by a mutex) using a new libdicom v1.2.0 function `dcm_filehandle_get_frame_number()` which allows fast readout of the internal libdicom tile index. Delaying index construction until use gives a very nice speedup for slide open. 
4. Small changes around testing the various UIDs which link files and levels together.

The new libdicom has two relevant changes:

1. The new `dcm_filehandle_get_frame_number()` as noted above. 
2. Support for the `ConcatenationFrameOffsetNumber` field, for `TILED_FULL` catenations.

Sample data:

I have some cc0 sample files kindly donated by Geneva University Hospital, and I think dclunie has many others. I'll upload one of them in a moment.

Geneva have been using this branch for a few weeks and say it's working well for them.

TODO:

- [x] update openslide to require libdicom 1.2.0, update build files, update wrap
- [x] fix failures in current tests
- [ ] update documentation
- [x] add new tests and sample data (maybe this should be a separate PR?)
- [ ] squash, sign and split PR (I'll do this at the end)


jcupitt:
Here's a sample file:

http://www.rollthepotato.net/~john/DS_A07R_01S.zip

5gb, level 4 is split into six parts, level 5 is split into two.

bgilbert:
Thanks for the PR!  Some preliminary things:

- By policy, I don't look at diffs until they're signed off.  I know you're good for it, but if the signoffs don't specifically need to be deferred, ideally you could rebase to drop the merge commits and sign off.
- We should probably have a separate preliminary PR to update the libdicom wrap and fix tests, due to ImagingDataCommons/libdicom@7381e3ea1907780458376162b41922260d82107a.  Automation would normally handle this, but it won't fix the tests of course.  Tests should use regexes to support both old and new libdicom.
-  Could you submit the sample data through the [official form](https://openslide.org/submit/) so we have a record of the license and details?

> Delaying index construction until use gives a very nice speedup for slide open.

Without having seen the code yet, I'll note that I'm skeptical.  We do have one case in the existing code (Hamamatsu VMS restart marker scan) which is deferred from open time because it's too I/O-intensive to do up front, but in that case the code runs incrementally in a background thread, and has a debug flag to force it to run eagerly.  It sounds as though this PR does the full work on the first `read_region`?  I'd rather not introduce some surprising tail latency.

> add new tests and sample data (maybe this should be a separate PR?)

Actually, adding in this PR is probably better, once the data goes into openslide-testdata.

jcupitt:
Here's another test file (15gb!!):

ftp://medical.nema.org/MEDICAL/Dicom/DataSets/WG26/3DHISTECH/FL_20191127/DICOM_TILED_SPARSE/MouseKidney_F24630_full

It's part of the 2023 connectathon, so it's recent, from the scanner manufacturer, and CC0. It doesn't work and probably should, I'll investigate.

jcupitt:
Thanks for the feedback!

I'll try to get libdicom 1.2.0 released in the next couple of days, then make a preliminary PR on openslide to use the new version.

I've chased the donor of the sample data to try to get the upload done.

I've removed the lazy index build, rebased and (this time) remembered to sign off the commit.

jcupitt:
For reference, I timed lazy index vs. index on open.

Right now (index on open) with a 6GB test scan (3Dhistech, levels split to stay within 500mb per file, no offset table), I see:

```
$ time openslide-show-properties 4_2 > /dev/null

real	0m4.527s
user	0m3.222s
sys	0m1.304s
```

If I put the lazy index back I see:

```
$ time openslide-show-properties 4_2 > /dev/null

real	0m0.016s
user	0m0.007s
sys	0m0.009s
```

And fetching a tile from level 4 with index on open:

```
$ time openslide-write-png 4_2 100 100 4 512 512 x.png

real	0m4.583s
user	0m3.333s
sys	0m1.249s
```

And with lazy index:

```
$ time openslide-write-png 4_2 100 100 4 512 512 x.png

real	0m0.048s
user	0m0.033s
sys	0m0.015s
```

Being able to get a low-ish resolution image on the screen quickly is really nice for interactive use.

jcupitt:
I've released libdicom v1.2.0 and made a preliminary PR to update openslide:

https://github.com/openslide/openslide/pull/658

Once that's merged I'll rebase this PR and remove the "draft" tag.

jcupitt:
All tests are now passing, I think.

We could add some more tests I guess, but maybe that should be in a followup PR?

I've not done a final squash yet. This seems like a fairly simple PR to me, so I'd probably make it a single commit. Is that OK?

choosehappy:
Would it be possible to bump the merge of this PR? 

we've been compiling and using it locally - but in our evolving CI/CD pipelines that local compilation is adding some complexity that would be removed post release of an official version  

we need this updated functionality for our clinical routine workflow. I can say that we've been using it internally for at least 4 months on ~1.5k slides per day without any issues - so functionality testing wise, seems rock solid!

Thanks

bgilbert:
Thanks for the experience report!  There's some infrastructure work I'm trying to finish before landing any more test cases, so I'm afraid it may be a bit, but I haven't lost track of this.  For your CI pipeline, could you build the necessary binaries once and stash them somewhere for reuse?

choosehappy:
great, thanks for the update.  happy to provide any more information from our side if useful

indeed - compiling and stashing is emerging as our to go method : ) this does create some additional complexities with versioning and share our tools externally with other hospitals, but we're managing. 

naturally a bit more elegant to have a public repo version we can just pull  and use directly ;-)

SebastianRiechert2:
+1 on the request to merge this. it would reduce the complexity of our docker builds, since this feature is the last dependency holdout we can't just install from conda/pypi

bgilbert:
Rebased.  I'll go through this and add some cleanups.

src/openslide-vendor-dicom.c
```diff
@@ -829,13 +902,32 @@ static bool maybe_add_file(openslide_t *osr,
...
+                "Unsupported image format");
+    return false;
+  }
+
+  int64_t value;
+  if (!get_tag_int(f->metadata, SamplesPerPixel, &value)) {
+      return false;
+  }
+  switch (value) {
+  case 1:
```
bgilbert: Was this change meant to be included in this PR?  If `SamplesPerPixel` is 1, Sup. 145 requires `PhotometricInterpretation` to be `MONOCHROME2`.  Since OpenSlide doesn't allow that, I don't think allowing `SamplesPerPixel == 1` here adds support for any additional slides.

bgilbert: I've removed this part for now.

bgilbert:
The changes were mostly refactoring and stricter metadata validation.  There was one substantive bug on well-formed datasets: if the `x` and `y` arguments to `openslide_read_region()` pointed to a sparse (missing) tile of a concatenated level, the call would not return any pixel data for the whole region.

bgilbert:
Thanks for implementing this!

PR #663, opened 2025-04-16
Drop support for Ubuntu 20.04 and Python 3.8

bgilbert:
After May 31, Ubuntu 20.04 will no longer be under standard security maintenance.

Python 3.8 is EOL and Ubuntu 20.04 is the last officially-supported distro that requires it.


bgilbert:
Actually, this doesn't need to wait for May 31.  There's no requirement that we support the `main` branch on Ubuntu 20.04 right up to EOL.

Issue #662, opened 2025-04-15
DICOM with mixture of transfer syntaxes and ANN files in folder won't open


dclunie:
### Operating system

macOS 13.6.6

### Platform

64-bit x86

### OpenSlide version

4.0.0.6-darwin-x86_64

### Slide format

DICOM

### Issue details

Tested openslide-4.0.0.6-darwin-x86_64.jar as used in QuPath-v0.6.0-rc4-Mac-x64.

QuPath using OpenSlide (BioFormats disabled) is unable to load from the folder of files, unless the ANN and two J2K (.90 Transfer Syntax) files are moved out of the folder, in which case the pyramid loads OK.

See files from DICOM WG 26 WSI Connectathon at this [Box location](https://hologic.ent.box.com/s/7bjeb78b4314hq2z1s6t7nda6ng1hm8j).

jcupitt:
The problem seems to be that this image uses lossless JP2K with the `PhotometricInterpretation` set to `YBR_RCT`, ie. reversible YCbCr. Currently, the openslide DICOM loader only supports RGB and `YBR_ICT` (irreversible YCbCr). I imagine the fix would be to add another member to `_openslide_jp2k_colorspace` for `OPENSLIDE_JP2K_YCBCR_REVERSIBLE`.

The equations are listed here:

https://dicom.innolitics.com/ciods/rt-dose/general-image/00880200/00280004

Or perhaps openjpeg has a path for this already we can use.

If I add something like this to the DICOM loader:

```C
    } else if (g_str_equal(photometric, "YBR_RCT")) {
      f->jp2k_colorspace = OPENSLIDE_JP2K_YCBCR;
      found = true;
    }
```

ie. load these jp2k images with the existing lossy YCbCr path everything else seems to work:

![Image](https://github.com/user-attachments/assets/7268d3da-dba3-4a20-a3e0-731b15cc21f0)

Though the colours are probably not correct.

jcupitt:
@dclunie would you mind posting a screenshot of roughly what that slide should look like? It'd be helpful while testing a PR.

jcupitt:
The slide loads well with libdicom and openslide using the small patch above, except for the hilariously wrong colours.

<img width="1807" height="1171" alt="Image" src="https://github.com/user-attachments/assets/853b2614-90e5-4bd2-9913-45eea8e877a8" />

I'll try making a PR that includes the correct colour transform.

dclunie:
[...]

PR #684, opened 2025-07-22
Add support for reversible JP2K DICOM (trivial patch)

jcupitt:
Lossless JP2K usually has `PhotometricInterpretation` set to `YBR_RCT` (reversible YCbCr). OpenJPEG does the conversion to RGB, so we can just use the RGB path in this case.

See https://github.com/openslide/openslide/issues/662


jcupitt:
The linked issue has a couple of sample images which load correctly with this PR.

dclunie:
What is the status of this?

bgilbert:
Rebased.

bgilbert:
Closing in favor of https://github.com/openslide/openslide/pull/743.  Thanks for looking into this.

PR #712, opened 2026-04-12
grok: add support for decompression of j2k tiles with this open source codec

boxerab:
Grok (https://github.com/GrokImageCompression/grok) is an AGPL licensed codec that provides full support for JPEG 2000 standard Part 1, and can outperform many commercial JPEG 2000 libraries.

It can be enabled at runtime with the environment variable

export OPENSLIDE_JP2K_BACKEND=grok

Tests have also been provided, all passing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,12 @@ To build OpenSlide, you will need:
 - zlib
 - Zstandard
 
+Optional:
+
+- [Grok][] ≥ 20.3 (alternative JPEG 2000 codec)
+
+[Grok]: https://github.com/GrokImageCompression/grok
+
 Then:
 
 ```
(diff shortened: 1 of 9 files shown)


bgilbert:
Thanks for the PR.  How are you using this code?  Does it provide any new functionality, or is this purely a performance improvement?

The licensing is an issue, of course.  OpenSlide is licensed under the LGPLv2.1, and a binary built with Grok support would be subject to the AGPLv3 even if the environment variable was not set.  As a result, Linux distributions would probably not enable the feature, nor would our own openslide-bin builds.  (We could ship a separate AGPL build but I'm doubtful anyone would use it, legally at least.)

It's [theoretically possible](https://www.gnu.org/licenses/gpl-faq.html#GPLPlugins) to produce an LGPL OpenSlide that talks to a separate Grok helper program using sockets or pipes, but there's a real question whether it's worth the effort.  Grok doesn't seem to be fully available in the ecosystem: Homebrew has it, Debian has it but not in `stable` or `testing`, and Fedora doesn't package it.  OpenSlide often favors simplicity over performance and tries to avoid optional dependencies, so we'd need a really substantial performance improvement to be worth the complexity.

boxerab:
@bgilbert thanks for your reply. Regarding performance, Grok can achieve up to 2X OpenJPEG decompression speed, especially for very large tiled images.

As you probably know, licensing issues only arise when distributing the software, so if one of your users wants to use OpenSlide + Grok for their individual use, AGPL vs LGPL is not relevant. Only if someone distributes OpenSlide + Grok with closed source changes would the AGPL forbid this aggregate product. So, as along as docs are clear about this, accepting this PR would simply allow your users more choice - not a bad thing IMHO.

Of course, it's up to you how to proceed.

PR #721, opened 2026-05-07
Simplify c.pre-commit job using prek-action

ulgens:
prek-action handles prek installation and cache management.

https://github.com/j178/prek-action


bgilbert:
Thanks for the PR.  The lesson from the old `pre-commit/action` was that the action code is simple enough that it's better to inline it rather than take the additional dependency.

PR #743, opened 2026-06-05
dicom: fix JPEG 2000 colorspaces

### LINKED ISSUE #526: OpenSlide does not support DICOM reversible (lossless) JPEG 2000 compressed images
### Operating system

macOS

### Platform

64-bit x86

### OpenSlide version

4.0.0

### Slide format

DICOM

### Issue details

Some large images that start out raw (uncompressed) and are encoded in DICOM, which has practical (2GB) and actual (4GB -2) limits on uncompressed pixel data size need to be compressed reversibly (losslessly) and this is done using JPEG 2000 reversible (integer wavelet and multi-component transform).

E.g., will not read this image, which is believed to be valid and is successfully handled by the BioFormats DICOM WSI reader as used in QuPath:

s5cmd --no-sign-request --endpoint-url https://s3.amazonaws.com cp 's3://idc-open-data/b2af35dd-265c-4e8c-8d0c-81cb28965a80/*' .

https://viewer.imaging.datacommons.cancer.gov/slim/studies/2.25.47398213943575571195567338623138358598/series/1.3.6.1.4.1.5962.99.1.2515301380.141547177.1686142514180.4.0

(These images are from IDC and are CC-BY-4.0).

### LINKED ISSUE #646: DICOM J2K wrong color with YBR_ICT and rejects YBR_RCT photometric interpretation
Tested openslide-4.0.0.3-darwin-x86_64.jar as used in QuPath-0.6.0-rc3-x64.

[ICT DICOM files are here](https://www.dropbox.com/scl/fo/onomtza9dti57i7evkcst/AF9olYXlL0Wo8DrsIQuNgTA?rlkey=qf55xbeqbqwlrbq371sasevbo)

[RCT DICOM files are here](https://www.dropbox.com/scl/fo/hrfs31c3jifptoj77sxmk/AHNQaCrBTZBiqx9Ulgi2Bqs?rlkey=0kyeux4ycvw12flh7hozw8i2p)

(these both load and display as expected with BioFormats selected rather than OpenSlide)

w. ICT shows wrong colors:

CMU-1_SVS_DICOM_j2klossy_ScreenShot_WrongColors

w. RCT says:

05:16:16.324	[JavaFX Application Thread]	ERROR	qupath.lib.images.servers.openslide.OpenslideServerBuilder	Unable to open file:/private/tmp/WSI_MultiFrame_j2klossless/img_0.dcm with OpenSlide: Reading /private/tmp/WSI_MultiFrame_j2klossless/img_0.dcm: Unsupported photometric interpretation YBR_RCT for 1.2.840.10008.1.2.4.90	java.io.IOException: Reading /private/tmp/WSI_MultiFrame_j2klossless/img_0.dcm: Unsupported photometric interpretation YBR_RCT for 1.2.840.10008.1.2.4.90
	at qupath.lib.images.servers.openslide.jna.OpenSlide.checkError(OpenSlide.java:147)
	at qupath.lib.images.servers.openslide.jna.OpenSlide.<init>(OpenSlide.java:103)
	at qupath.lib.images.servers.openslide.jna.OpenSlideLoader.openImage(OpenSlideLoader.java:141)
	at qupath.lib.images.servers.openslide.OpenslideImageServer.<init>(OpenslideImageServer.java:144)
	at qupath.lib.images.servers.openslide.OpenslideServerBuilder.buildServer(OpenslideServerBuilder.java:58)
	at qupath.lib.images.servers.ImageServerBuilder$DefaultImageServerBuilder.buildOriginal(ImageServerBuilder.java:366)
	at qupath.lib.images.servers.ImageServerBuilder$AbstractServerBuilder.build(ImageServerBuilder.java:190)
	at qupath.lib.gui.QuPathGUI.openImage(QuPathGUI.java:1722)

### LINKED ISSUE #662: DICOM with mixture of transfer syntaxes and ANN files in folder won't open

bgilbert:
[CMU-1-JP2K-33005.zip](https://openslide.cs.cmu.edu/download/openslide-testdata/DICOM/CMU-1-JP2K-33005.zip) has a `PhotometricInterpretation` of `RGB` but is actually `YBR_ICT`.  [JP2K-33003-1.zip](https://openslide.cs.cmu.edu/download/openslide-testdata/DICOM/JP2K-33003-1.zip) has a `PhotometricInterpretation` of `YBR_ICT` but is actually `YBR_FULL_422`.  We shouldn't base our behavior on those samples.

Add support for `YBR_RCT` and correct the colorspace of `YBR_ICT`.  Both of these are transparently decoded to RGB by OpenJPEG.

Don't add support for `YBR_FULL_422`, which no one has actually requested.  Note 3 of [DICOM PS3.5 2026b section 8.2.4](https://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_8.2.4.html) suggests it would be permissible, but it's not listed in Table 8.2.4-1 so may not exist in the wild.  As a result, JP2K-33003-1.zip will not decode correctly.

Replace the test cases that used the invalid sample data.

Also forbid `YBR_ICT` with the lossless JP2K transfer syntax, because the standard does.

Fixes: https://github.com/openslide/openslide/issues/526
Fixes: https://github.com/openslide/openslide/issues/646
Fixes: https://github.com/openslide/openslide/issues/662

cc @jcupitt


jcupitt:
Looks great! 

I suppose my only concern would be that this kind of confusion (container metadata does not match contained object properties) happens constantly and everywhere :( TIFF is notorious for exactly this kind of out of spec foul up, for example.

I think the usual advice with crazy^Wflexible formats like this is pedantic-on-write, permissive-on-read. If openslide only reads in spec images, it's probably going to fail on at least some real world cases (though I don't have an example handy).

bgilbert:
Yeah, it's a recurring problem for OpenSlide.  As I said in a test case comment, though, I believe [Postel](https://en.wikipedia.org/wiki/Robustness_principle) was [wrong](https://datatracker.ietf.org/doc/rfc9413/).  In general I prefer to be pedantic, then wait for a bug report.  Otherwise it's very difficult to know what format variants we actually need to support, which makes it hard to extend or refactor existing code.

(That means OpenSlide won't stay relevant unless bug reports get addressed, which unfortunately I haven't been consistent about.  I think the principle is still sound in any case.)

If there are slides like JP2K-33003-1.zip which claim `YBR_ICT` but need YCbCr conversion, we'd have to try to detect that from chroma subsampling.  I'm doubtful there are non-synthetic slides that need this, so I'd rather not add that code just yet.

Thanks for taking a look!

PR #745, opened 2026-06-05
dicom: preload missing tiles in sparse levels

bgilbert:
Search for them in advance rather than discovering them at runtime.  This will cause libdicom to load level offset tables earlier, which is slower during open but avoids tail latency the first time we read a tile.  Once we have the new libdicom `find_frame_number` API, it'll also avoid `DcmError` allocation and string-formatting overhead from trying to read sparse frames with `dcm_filehandle_read_frame_position()`.

cc @jcupitt

### DIFF
--- a/src/openslide-vendor-dicom.c
+++ b/src/openslide-vendor-dicom.c
@@ -121,6 +121,7 @@ static const char BitsAllocated[] = "BitsAllocated";
 static const char BitsStored[] = "BitsStored";
 static const char Columns[] = "Columns";
 static const char ConcatenationUID[] = "ConcatenationUID";
+static const char DimensionOrganizationType[] = "DimensionOrganizationType";
 static const char HighBit[] = "HighBit";
 static const char ICCProfile[] = "ICCProfile";
 static const char ImageType[] = "ImageType";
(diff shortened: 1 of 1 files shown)


jcupitt:
I timed the effect on startup costs and it seems pretty small now -- using  `dcm_filehandle_find_frame_number()` as well it only ends up about about 20% more on this PC with a big sparse DICOM. So it's a thumbs up here.

The seek cache seems to help a lot.

jcupitt:
The patch I used for find_frame_number, fwiw:

```patch
diff --git a/src/openslide-vendor-dicom.c b/src/openslide-vendor-dicom.c
index 35f1d18d..d53705b3 100644
--- a/src/openslide-vendor-dicom.c
+++ b/src/openslide-vendor-dicom.c
@@ -1126,20 +1126,19 @@ static bool finalize_level(struct dicom_level *l, GPtrArray *files,
         for (file_num = 0; file_num < l->file_count; file_num++) {
           DcmError *dcm_error = NULL;
           uint32_t n;
-          if (dcm_filehandle_get_frame_number(&dcm_error,
-                                              l->files[file_num]->filehandle,
-                                              tile_col, tile_row, &n)) {
-            // found one, record the file that has this tile and break
-            if (l->tile_files) {
-              l->tile_files[tile_index] = file_num;
-            }
-            break;
-          } else if (dcm_error_get_code(dcm_error) == DCM_ERROR_CODE_MISSING_FRAME) {
-            dcm_error_clear(&dcm_error);
-          } else {
+
+          if (!dcm_filehandle_find_frame_number(&dcm_error,
+                                                l->files[file_num]->filehandle,
+                                                tile_col, tile_row, &n)) {
             _openslide_dicom_propagate_error(err, dcm_error);
             return false;
           }
+
+          if (n != 0 && l->tile_files) {
+            // found one, record the file that has this tile and break
+            l->tile_files[tile_index] = file_num;
+            break;
+          }
         }
         if (file_num == l->file_count) {
           // none
```

bgilbert:
Great, thanks for checking!

(For the record, that patch marks every tile in a non-concatenation `TILED_SPARSE` level as sparse, but that wouldn't have affected your performance numbers.)

jcupitt:
I tested with a sparse DICOM with levels split across files, in case I wasn't clear.

bgilbert:
This PR only affects non-split levels (split ones were already preloading) so in that case your performance numbers wouldn't show the impact on level 0.

jcupitt:
Ah gotcha. I've not looked at the openslide dicom loader in detail for a while :(

PR #755, opened 2026-06-08
util: replace GRegex preprocessing with manual scan

kleisauke:
Replace GRegex-based preprocessing with a simple line scan using
g_ascii_isspace() and pointer arithmetic. This removes regex overhead
while preserving behavior.

No functional change.

### DIFF
--- a/src/openslide-util.c
+++ b/src/openslide-util.c
@@ -108,45 +108,35 @@ GKeyFile *_openslide_read_key_file(const char *filename, int32_t max_size,
 
   // remove empty keys reported to occur in MIRAX Slidedat.ini
   if (flavor == OPENSLIDE_KEY_FILE_MIRAX) {
-    static gsize have_re = 0;
-    static GRegex *re = NULL;
-    // g_once_init_enter_pointer() on glib 2.80+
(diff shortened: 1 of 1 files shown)


kleisauke:
Context: this is useful for libvips its Windows builds, where GRegex is removed pre-emptively, see:
https://github.com/libvips/build-win64-mxe/blob/v8.18.3-rc1/build/patches/glib-2-without-gregex.patch

kleisauke:
See also PR https://github.com/lovell/sharp-libvips/pull/106 for context on why libvips applied the patch mentioned above. However, it appears that GLib has since replaced pcre1 with pcre2.

bgilbert:
Yeah, this is probably better.  I went with GRegex thinking it was going to handle the string replacement too, but that didn't work out.  And I wouldn't mind removing PCRE2 from openslide-bin too.  Will take a look, thanks.

bgilbert:
I've implemented this a little differently in https://github.com/openslide/openslide/pull/762, doing the entire scan in one pass.  Thanks for the PR, and for pointing this out!

kleisauke:
Thank you!

